PR #27, opened 2017-03-01
Make a proper Readme

licaon-kter:
with F-Droid link and all

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,27 @@
+# Orgzly is an outliner for notes and tasks
+
+[<img src="https://f-droid.org/badge/get-it-on.png" alt="Get it on F-Droid" height="60">](https://f-droid.org/app/com.orgzly)<a href="https://play.google.com/store/apps/details?id=com.orgzly"><img src="https://play.google.com/intl/en_us/badges/images/generic/en_badge_web_generic.png" height="60"></a>
+
+## Features:
+
+* It stores notebooks in plain-text and keeps them where you choose to
+* Create, edit and delete notes and tasks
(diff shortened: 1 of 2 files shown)


nevenz:
Can we:
* Use `README.org` instead of Markdown, it's just feels weird not too. :)
* Fix first bullet to say "Store notebooks... and keep them ..."
* Put Google Play badge before F-Droid. I prefer users using the app on Google Play if the can.

Thanks!

gdt:
It seems strange to prefer google play, since org is part of the Free Software world and I would expect org things to have a bias to Free network services over proprietary services.

nevenz:
I (as a developer) prefer Google Play for the stats and more importantly crash reports.

licaon-kter:
> Use README.org instead of Markdown, it's just feels weird not too. :)

Feel free to edit it, I don't know how (org sintax). ;)
Be sure to test it though, the previous one had the `#+html:` thing that was useless as a webpage (eg. use ForkHub to read that file). Markdown is kinda standard around GitHub.

> Fix first bullet to say "Store notebooks... and keep them ..."

Right, sorry.

> Put Google Play badge before F-Droid. I prefer users using the app on Google Play if the can.

Done, although you realise F-Droid users won't start using the Google Play store one or the other way around based on logo position, right? :)

nevenz:
`#+html:` should work (I had to use that to make image smaller, as `#+ATTR_HTML` doesn't seem to be supported by GitHub), though I agree Markdown makes that easier.

If someone wants to play with `README.org` in the future, great. :)

Thanks again!

PR #29, opened 2017-03-02
Typo fix

licaon-kter:
...do fix it in the Google Store description too


nevenz:
Oops, fixed in store too, thanks.

PR #39, opened 2017-03-12
Create "Share notebook" setting.

sarg:
Hi, there.
I am employing GTD system and would like to put all new items in "refile" notebook. 
I have created new setting "share notebook" for such a case.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/EspressoUtils.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/EspressoUtils.java
@@ -48,17 +48,17 @@ import static org.hamcrest.Matchers.not;
 class EspressoUtils {
     static final int SETTINGS_REVERSED_NOTE_CLICK_ACTION = 1;
 
-    static final int SETTINGS_STATE_KEYWORDS = 19;
-    static final int SETTINGS_DEFAULT_PRIORITY = 20;
-    static final int SETTINGS_LOWEST_PRIORITY = 21;
+    static final int SETTINGS_STATE_KEYWORDS = 20;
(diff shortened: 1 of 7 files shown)


nevenz:
Great, thanks!

Few random notes:
* Indexes in `EspressoUtils` need updating, as tests will not pass now
* `title` needs extracting to `strings.xml`, so it can be translated
* We should rename default notebook (now `Share`). All this is now also used when creating new note from the ongoing notification, not just when sharing to app. I should have done that before when that feature was implemented. I think `Inbox` would be best, it's kinda standard and it's even GTD. :)

It would be great if user would be able to select one of the existing notebooks (`ListPreferenceWithValueAsSummary` populated dynamically) instead of having to type the name, though I realize that's a bit more work and I'm fine with this being `EditTextPreferenceWithValueAsSummary` for now.

Cool about `selectAllOnFocus`, I haven't seen that one before, it should probably be added in quite a few places where it makes sense. Speeds up things a lot.

sarg:
I have updated my PR with your proposed fixes.
BTW, when I run test suite on my phone, some of the tests fail:

```
android.support.test.espresso.PerformException: Error performing 'single click' on view 'Animations or transitions are enabled on the target device.
For more info check: http://goo.gl/qVu1yV

with text: is "Delete"'.
```

It will be nice if you document how do you build and test the application.

nevenz:
You have to turn off animations on the device as per that URL. I don't think there's anything app-specific for running the tests, though I don't mind updated README with instructions.

Only `testDefaultBookRemainsSetAfterRotation` and `testDefaultBookName` are failing for me, as they're still using `Share` notebook name.

PR #66, opened 2017-04-17
Added the ability to create shortcuts on the homescreen, fixes #46

### LINKED ISSUE #46: Widget to open a specific notebook
See for example [OI Shopping list](http://www.openintents.org/shoppinglist/) as an application that supports this. I use this feature in OI shopping list to have a number of different checklists that I can open with one click.

I fondly imagine that being able quickly open checklists like this and organize them on a screen makes one more likely to use them, as it reduces the cognitive load of having to click through some tasks.

One might also want to be able to jump to a particular node with a tree.

MackieLoeffel:
This PR adds the ability to add a shortcut to a specific notebook to the homescreen, as was requested in #46.

The code is almost complete, but the last time, when I wrote code for Android, was quite a few years ago, and there are some open points, which need to be addressed:
- [x] What is the best way to get the title of a book from its id? I need it to set the title of the shortcut.
- [x] I added a new string, does it need to be translated to all languages?

@nevenz Can you please look at this PR and give me some feedback, what should be fixed, before you can merge this?


nevenz:
You can get `Book` with `new Shelf(context).getBook(bookId)`. Probably should check for null, just in case (of running sync deleting the book - might not be possible currently, but in the future).

I noticed the created icon is larger then app's existing icon, which looks weird.

Thanks!

MackieLoeffel:
Thanks, I adjusted the title of the shortcut. I've tried quite a bit to make the Shortcut icon the same size as the normal icon, but couldn't get it to work. I found this http://stackoverflow.com/questions/14084259/get-correct-laucher-icon-size-on-android and added scaling to the correct size, but it doesn't make a difference, because the launcher always scales the icon of the shortcut to a slightly bigger size than the normal icon. Do you have any ideas, what the problem could be?

nevenz:
> I've tried quite a bit to make the Shortcut icon the same size as the normal icon, but couldn't get it to work.

It might not be the issue with shortcut icon being too big, but launcher icon being too small. It looks like it's being resized if it reaches a certain size.

For example, this doesn't happen if `cic_dropbox_black_48px` or `ic_folder_open_black_48dp` are used instead (replacing current resource in manifest and activity).

Also, if I open `cic_orgzly_logo.png` (for example from `drawable-xxxhdpi`, deleting logos from other directories to make sure it's the one that will be used) and then simply erase the edges a bit to make the content occupy less space (not changing the actual size of the image), then it's also not being resized and both icons are the same.

No idea what's going on there and if it's a bug or feature.

If we don't figure this out, perhaps a different icon (logo changed slightly to represent a notebook - I'll play a bit with that) would be OK. It's still going to be larger, but it shouldn't be looking that weird if it's a different icon completely.

BTW, can we:
1. Name the widget "Notebook" (I think it's matter of "label" in activity). There should be a string resource that can be used.
2. Shorten or remove completely the title when choosing notebook. It's too much text now. I vote for removing it altogether - if user got there, he should know what he's going at that point. And if not, he'll remember soon enough after clicking the notebook. 😄

MackieLoeffel:
About the logo: Strange, I have no idea, what is happening with the resizing. A new logo would be great, because it would be easier to differentiate between a shortcut and the normal orgzly link. But I'm not very good in creating well looking logos. It would be great, if you could design something (maybe the orgzly-logo with a small notebook at the bottom right? Just an idea.)

About your two points:
1. I renamed it, but there was no string resource with the text "Notebook" yet. So I created a new one.
2. Yes, the title is quite long, what do you think about only "Pick a notebook"? Or I can remove it completely, whatever you prefer.

Is there anything else, which should be fixed before merging?

nevenz:
> maybe the orgzly-logo with a small notebook at the bottom right?

Yup, I was thinking something like that.

> Yes, the title is quite long, what do you think about only "Pick a notebook"?

Sure. It does look strange now without the title. 😄

> Is there anything else, which should be fixed before merging?

Don't think so.  Maybe you can just remove the scaling, since it doesn't work anyway.

Thanks!

MackieLoeffel:
> Sure. It does look strange now without the title. 😄

Title is back in

> Don't think so. Maybe you can just remove the scaling, since it doesn't work anyway.

Done

nevenz:
Cool.

I've pushed new `cic_orgzly_logo_with_notebook` icon (and modified the original, leaving out bullet points to make some space 😄).

MackieLoeffel:
Great, i adjusted the code to use the new logo for the shortcut.

nevenz:
It would be cool if the logo with notebook is also displayed in the widget list when picking the shortcut. But that might require messing with the widget API, which if it's the case, is probably not worth it.

MackieLoeffel:
Good idea, the icon in the widget picker list is now the new one with the notebook. It was only a small adjustment in the manifest.

talwrii:
Awesome. I'm excited about the prospect of this feature going live and I am open to doing some beta testing of it.

PR #72, opened 2017-05-01
Add option to disable separation of header and content with an empty line

### LINKED ISSUE #17: Add option to disable separation of content from heading
As discussed in #12 it would be nice to have an option to control the whitespace between the heading and content of a note similar to the option `Separate notes with an empty line`. Personally, I prefer to only have a blank line between headings and multiline content.

MackieLoeffel:
Fixes #17. Depends on orgzly/org-java#2.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/EspressoUtils.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/EspressoUtils.java
@@ -12,30 +12,18 @@ import android.support.test.rule.ActivityTestRule;
 import android.view.View;
 import android.widget.ListView;
 import android.widget.Spinner;
-
 import com.orgzly.R;
-
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeMatcher;
 
-import static android.support.test.espresso.Espresso.onData;
-import static android.support.test.espresso.Espresso.onView;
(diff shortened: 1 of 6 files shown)


nevenz:
Positions in `EspressoUtils` just need to be updated since new preference was added before few of them.

Thanks!

MackieLoeffel:
I updated the positions. Is there anything else?

And I have to admit, that I usually don't run all the tests, because they take really long. I even tried to setup an automatic build on travis, but it exceeded the timelimit of 50 minutes ( you can see it here https://travis-ci.org/MackieLoeffel/orgzly-android/builds/227367722).

nevenz:
> And I have to admit, that I usually don't run all the tests, because they take really long. I even tried to setup an automatic build on travis, but it exceeded the timelimit of 50 minutes ( you can see it here https://travis-ci.org/MackieLoeffel/orgzly-android/builds/227367722).

Hmm, they take 12min on my end currently.

Is it possible they get blocked due to external storage permission prompt, which does appear if permission is not granted before?

MackieLoeffel:
> Hmm, they take 12min on my end currently.

Hm, ok, they take a little less than 30 minutes on my laptop.

> Is it possible they get blocked due to external storage permission prompt, which does appear if permission is not granted before?

That could be, whats happening. I will take a look, whether this is the problem. But I don't have much hope, because the android support on travis is not very good and it took me quite a while to get this far.

PR #83, opened 2017-05-15
Open orgzly from widget

MackieLoeffel:
Add the ability to open orgzly directly from the search widget by clicking on the icon.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/uiautomator/ListWidgetTest.java
+++ b/app/src/androidTest/java/com/orgzly/android/uiautomator/ListWidgetTest.java
@@ -21,13 +21,17 @@ import static org.hamcrest.Matchers.notNullValue;
  * tested on an Nexus 5 Emulator with Android 25
  * Must have an instance of the ListWidget on the Homescreen, when starting this test
  * Also the default version of Getting Started with Orgzly must be the only notebook
+ *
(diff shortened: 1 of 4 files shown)


MackieLoeffel:
What do you mean with "overview button"? An Orgzly shortcut on the homescreen?

I use the widget at full size on a single homescreen and it would be nice to be able to open orgzly from there without opening a note. (And I think it would be nice, if the rest of the title bar was also clickable and would open the search selection, maybe I will add this to this PR).

nevenz:
No, I mean instead of this:

```
        /* bring orgzly back to front, see http://stackoverflow.com/a/5502950/7757713 */
```

open Orgzly and show results for the saved search that's selected in the widget.

Currently if I leave Orgzly while I'm viewing some note, do something else, then go to the widget and click the icon, I'm brought back to the note. Which is not what I would expect by clicking the icon which is right next to the saved search on the widget.

Overview button is the square on Android  in bottom right corner that opens the list of apps. If I want the above behaviour (that's what I mean by "resuming work"), I would open Orgzly using that button and selecting it from the list.

> And I think it would be nice, if the rest of the title bar was also clickable and would open the search selection, maybe I will add this to this PR

Sounds good.

MackieLoeffel:
Ok, now I understand what you mean and it sounds good. I will adjust it.

MackieLoeffel:
I updated the PR with the changes we decided.

Issue #10, opened 2017-02-17
Fold/unfold all doesn't work?


alphapapa:
Hi,

I just installed and started using Orgzly with my existing Org notes in Dropbox.  It seems to work great, but the fold/unfold-all button doesn't seem to do anything.  No matter how many times I tap it, or how long I wait, or how small a file I try it on, it doesn't do anything--so I have to collapse everything manually, which is obviously a chore in my large Org files.  :)

Thanks for your work!

nevenz:
Yeah, this is a big one and main reason why #8 (auto sync) isn't yet implemented.

It's not just fold/unfold, anything that needs to write to db will be queued.

PR #87, opened 2017-05-16
Sync after new note is created

psamim:
Hi,

I added the the option to sync after a new note is created. 

It works but sometimes it has a problem. As I really want to start understanding the project I would really appreciate if someone would take a look and tell me what is wrong. Sometimes after new note creation, `BooksClient.isModifiedAfterLastSync()` is still true therefore the out-of-date indicator besides notebook is visible. It only happens on my emulator while it works fine on my actual phone.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/EspressoUtils.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/EspressoUtils.java
@@ -45,8 +45,10 @@ class EspressoUtils {
 
     static final int SETTINGS_REPOS = 33;
 
-    static final int IMPORT_GETTING_STARTED = 38;
-    static final int SETTINGS_CLEAR_DATABASE = 39;
+    static final int SETTINGS_SYNC_AFTER = 34;
+
+    static final int IMPORT_GETTING_STARTED = 39;
+    static final int SETTINGS_CLEAR_DATABASE = 40;
 
     /**
      */
--- a/app/src/androidTest/java/com/orgzly/android/espresso/SyncingTest.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/SyncingTest.java
@@ -26,6 +26,7 @@ import static android.support.test.espresso.contrib.DrawerActions.open;
 import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
 import static android.support.test.espresso.matcher.ViewMatchers.withText;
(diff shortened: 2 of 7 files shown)


psamim:
This feature was discussed in #8

MackieLoeffel:
>  Sometimes after new note creation, BooksClient.isModifiedAfterLastSync() is still true therefore the out-of-date indicator besides notebook is visible. It only happens on my emulator while it works fine on my actual phone.

I had a similar bug today as well (without your code): A notebook had always the modified indicator, no matter how often I synced. I could only remove it by creating a conflict and resolving it. I don't know, if this is the same bug.

nevenz:
I have an old TODO about that, with only "When copying file to/from Dropbox?" as a note. So I've seen the issue before too.

nevenz:
It's not related though, I've checked the code - `onNoteCreateRequest` is async, so sync requested right after it is not guaranteed to run *after* the note has been created. Not fully at least (as book does end up being saved).

I think you want sync started in `onNoteCreated` (callback in `MainActivity`).

And we probably want some tests created for this. Perhaps a test in `SyncingTest` that would check that the new note exist and that there's nothing to sync (no sync icon for the notebook).

psamim:
I changed the code and now start sync in [`onPostExectue`](https://github.com/orgzly/orgzly-android/pull/87/files#diff-afabfc52a99f255dd6bfd3126032cbfa) of `createNote` async task before calling the callbacks. This way it works in `MainActivity` and `ShareActivity`.

I will add some test. Thanks for feedback.

psamim:
I added a passing test too. Please take a look. Thanks.

By the way I have multiple failing tests in master branch.

nevenz:
> I added a passing test too. Please take a look.

Thanks, looks fine to me.

BTW I'll probably have to either add a big fat warning for this option, or move it to a kind of "Advanced options" screen (I'm not a big fan of those, but might be a good solution).

Same reasons auto-sync hasn't been implemented yet:
- Generating file from DB data is slow for larger notebooks
- DB can be blocked during sync, so changes would lag (#10)
- Complete notebooks (files) are being sent over the net
- Specific to this feature: sync won't be started if it's already running

Either of this can either confuse an average user or/and piss him off.

> By the way I have multiple failing tests in master branch.

Which ones?

New ones for widget need some setting up (perhaps we should turn them off by default), as well as few that are writing to external storage, if you haven't granted the permission previously.

psamim:
Thanks for review and merge.

I agree with warning or Advanced options. 

It works in my case so maybe it works for someone else too. My sync repo is a local directory (Syncthing) and I keep notebooks small by regularly archiving old tasks using org-mode Emacs.

PR #96, opened 2017-05-29
Add sync action to new note notification and add sync notifications

### LINKED ISSUE #89: Add a sync action to the ongoing notification
Idea from: https://github.com/orgzly/orgzly-android/issues/85#issuecomment-301975036

The idea is to add an action to the ongoing notification, which starts the synchronisation.  
It would also be great, if a notfication is displayed during the synchronisation to see the status (only if not started from the app?).

I will work on this, when I find the time.

MackieLoeffel:
Fixes #89.

This PR adds a sync action to the ongoing notification and adds two new notifications, related to sync:
- the first is displayed while the sync is running
- the second is displayed, if an error occurred during sync or there are conflicts

The notifications have to be enabled in the settings.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/EspressoUtils.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/EspressoUtils.java
@@ -47,8 +47,8 @@ class EspressoUtils {
 
     static final int SETTINGS_SYNC_AFTER = 34;
 
-    static final int IMPORT_GETTING_STARTED = 39;
-    static final int SETTINGS_CLEAR_DATABASE = 40;
+    static final int IMPORT_GETTING_STARTED = 40;
+    static final int SETTINGS_CLEAR_DATABASE = 41;
 
     /**
      */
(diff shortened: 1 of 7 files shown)


nevenz:
Cool, thanks!

I've noted these (things I think we should do - let me know if something doesn't make sense) for now after trying it out:

- [ ] Replace sync icon with logo (current icon suggests that something is in progress)
- [ ] Add button for "New note", now that we have "Sync" (they are both actions, should be equal)
- [ ] Set content intent to just open the app (after the above)
- [ ] Use `startForeground` for sync service, if ongoing notification should be displayed while syncing is in progress
- [ ] Click on error notification to open the app

MackieLoeffel:
>  Replace sync icon with logo (current icon suggests that something is in progress)

I replaced the icon for the error notification. I think the sync icon makes sense for the sync in progress notification, because it is only shown, if the sync is in progress.

> - [ ] Add button for "New note", now that we have "Sync" (they are both actions, should be equal)
> - [ ] Set content intent to just open the app (after the above)  
>Update: I'm not sure about these last 2 actually, because when notification is not expanded, action buttons are not visible, and new-note should be a very quick action (no additional notification expanding).

Yes, the need to expand the note was the reason, why I left the new note action where it is.

>  Use startForeground for sync service, if ongoing notification should be displayed while syncing is in progress

Done.

>  Click on error notification to open the app
 
Done.

PR #120, opened 2017-06-29
Sorting by note creation date

### LINKED ISSUE #114: Search queries: Sort by creation date
I think it would be a nice quality-of-life improvement to be able to sort by the creation date of the note, so that the newer notes would pop up at the beginning of the search (since they're more likely to be more relevant if they were created recently).

cmdd:
With this pull request, **notes can now be sorted by their creation time** by using `o.c`, `o.create`, or `o.created` in search.

Internally, the created property (as defined in the settings) gets added to the database alongside the scheduled time, deadline time, etc.

When the created keyword is updated in the settings, the corresponding property name in the notes is _not_ updated. For example, if a note has a `CREATED` property and the created keyword is changed from `CREATED` to `CTIME` in the settings, the `CREATED` property of the note will not be changed to `CTIME`, and if the note doesn't already have a `CTIME` property, the note will not have a created time in the database.

This pull request fixes #114.


nevenz:
Can we use just an integer (unix timestamp) in `notes` table instead (`created_at`)? It seems simpler.

And I think it should be updated always, and not depend on any property at all, but be an independent value.

And then if "Created at" property is set, that property could be added for each note for which we have `created_at` value in `notes`. And when notes are imported to DB from a file, `created_at` can be set from the defined property.

Basically decoupling the two - "Created at" property is only useful when importing/exporting notebooks to Org format, as a way to keep `created_at`. But `created_at` and sorting by it is relevant even if the user never uses the Org format.

Does that make sense?

BTW, I get a crash with older version already installed. Table needs to be modified in `DatabaseMigration`.

Thanks!

cmdd:
I've updated my branch in regards to your comments:
- Database structure has been simplified, so now I've only added two fields to the `notes` table: `created_at` and `created_at_internal`
  - The purpose of `created_at_internal` is to maintain a record of the note's creation date independent of any created property.
  - When a note is created or imported from a file, both `created_at` and `created_at_internal` are set to the property with the created keyword, or the current date if that property doesn't exist.
  - When a note is updated or the created keyword is updated:
    - if the property with the created keyword of a note no longer exists, the `created_at` field will be set to the value of `created_at_internal`.
    - if a note did not have a property with the created keyword before but has one after the update, the `created_at` field will be updated, but the `created_at_internal` **will not**.
  - By doing this, Orgzly can remember the "original" creation date of a note, and can use that when the created property is changed.
- I've updated the `DatabaseMigrations` class with the required fields. I'm not sure why it causes a merge conflict but oh well!

nevenz:
> When a note is created or imported from a file, both created_at and created_at_internal are set to the property with the created keyword, **or the current date if that property doesn't exist**.

Do we really want this? Import time is not note's creation time - that remains unknown and we should perhaps keep that fact?

I think `Note` should have a `createdAt`. There's some DB querying now which we shouldn't do (like inside `NotesClient#update`). I'm still not sure we really need two values, but I might be missing something.

Anyway, I think we need some tests for this, to cover all the cases, as this now seems more complicated then I though it would be.

What would help test this is displaying note's `createdAt` in note fragment.

Even if we end up needing two values, only one really matters to the user - the one that will be displayed and used for sorting.

I think we should also change the settings - both enabling the property and changing its value should probably trigger `createdAt` updating, which currently doesn't happen. Perhaps a single spinner would be easier. "Insert note's creation time" could be "Property for note's creation time". As it's a property which should be kept in sync with `createdAt`.

Back to tests, something along these lines:

```
     /*
         * Create new note
         *
         * Check: note has created_at field set
         */

        /*
         * Change setting to use "CREATED" for created_at
         * Create new note
         *
         * Check: note has CREATED property set
         */

        /*
         * Create new note
         * Change setting to use "CREATED" for created_at
         *
         * Check: note has CREATED property set
         */

        /*
         * Create rook with a note that has CREATED property
         * Sync
         *
         * Check: note has CREATED property set
         * Check: note doesn't have created_at set
         */

        /*
         * Create rook with a note that has CREATED property
         * Sync
         * Change setting to use "CREATED" for created_at
         *
         * Check: note has CREATED property set
         * Check: note has created_at set
         */

        /*
         * Create rook with a note that has CREATED property
         * Change setting to use "ADDED" for created_at
         * Create new note
         * Sync
         *
         * Check: note one has CREATED property set and has *no* created_at time set
         * Check: note two has ADDED property set and *has* created_at time set (same value)
         *
         * Change setting to use "CREATED" for created_at
         *
         * Check: note one has CREATED property set and *has* created_at time set (same value)
         * Check: note two has ADDED property set and *has* created_at time set (same value)
         */

        /*
         * Create rook with a note that has CREATED property
         * Change setting to use "CREATED" for created_at
         * Sync
         *
         * Check: note has CREATED property set and *has* created_at time set (same value)
         *
         * Change setting to use "ADDED" for created_at
         * Create new note
         *
         * Check: note one *has* CREATED property set and *has* created_at time set (same value)
         * Check: note two *has* ADDED property set and *has* created_at time set (same value)
         */
```

I can see to merge this work so far to a new branch and we can continue there.

And thanks again for starting this and working on it.

nevenz:
I've merged this and created https://github.com/orgzly/orgzly-android/tree/feat/created-at

cmdd:
The benefit of having two values is that after creating a note, if you set its created property and then unset it, you can retain the original created time, rather than the one marked by the created property which no longer exists.

nevenz:
> after creating a note, if you set its created property

If option is enabled after creating note, I think then `createdAt` (potentially one and only) should be set (overwritten) to the value of the property if that property exists. If it doesn't exist, property would be created with `createdAt` value.

When property name is changed in settings, same thing would happen - set new property to `createdAt` or use its value instead if it exists. I do think it's fine to use the value of previous property, as that was `createdAt`'s "vessel" at the time. A valid property, which user wanted to rename.

At least I think that's fine, I'm curious about real usage examples which wouldn't be possible with this kind of implementation.

nevenz:
Sorry for the delay, I was postponing this one as I wasn't sure about the whole property vs internal value thing.

Perhaps this should just be ordering using the property value for now to make it simple. Anyway, I'll go through it all again soon.

I've deleted the old `feat/created-at` branch BTW.

cmdd:
No problem! I've been a little busy myself...

My personal branch only uses the property value right now (I removed the internal value), so I'll try to get all the upstream changes in my branch and add any necessary tests as soon as possible.

cmdd:
@nevenz I've updated my branch to work with current `master` branch, and I've added a few tests to make sure the created_at functionality is working correctly. Can this PR be merged in its current state?

app/src/main/java/com/orgzly/android/query/user/BasicQueryParser.kt
```diff
@@ -40,7 +40,8 @@ open class BasicQueryParser : QueryParser() {
             }),
 
             // scheduled:<3d
-            ConditionMatch("""^(scheduled|deadline|closed):(?:(!=|<|<=|>|>=))?(.+)""", { match ->
+       val intervalMatch = matcher.group(3)
```
nevenz: This is causing compilation to fail.

nevenz:
I though we're only going to use a note property for now (no new db field, no re-parsing notes etc.)?

Currently, used field is not set when notebook is loaded. You have to force the re-parse (in any way) for it to be set.

nevenz:
So querying and sorting by properties aside (which we want to be able to do with any properties eventually)...

I think the main issue of getting this right (and me finally wrapping my head around it), is the tight coupling of properties with created-at time. There are too many scenarios and ways to get weird results (I've started to list them all and it quickly went out of control).

This is my fault for allowing "Created at" property to be inserted in the first place, without thinking about created-at time as something independent, which always exists.

Anyway, here's a new proposal:

- [ ] Add `created_at` field to `notes` table as a simple `INTEGER`.

I think this is the most important part and everything else should be built around it. It's what's going to be used for searching and sorting. I think it used to be `INTEGER` early in PR, I'm not sure at what point it changed. It could have been me adding more confusion.

- [ ] Make `created_at` visible in note fragment and/or list view

Not essential, but it would help debugging and make things a lot clearer. Being able to see that main `created_at` next to property values will be of great help.

I'd also change settings:

- [ ] Remove "Created at" from "Notes & Notebooks"

As mentioned, created-at time should *always* be recorded, but not as a property.

Storing created-at time in a property is Org-specific and I'd move those few preferences there:

- [ ] Remove "Created at property" from "Notes & Notebooks"

- [ ] Add "Created-at property name" to "Org file format" (`CREATED` value)

- [ ] Add "Load created-at time from property" to "Org file format" (action, similar to re-parsing notes on settings changes, but explicit)

- [ ] Add "Write created-at time to property" (switch)

I think all these will make logic more straightforward.

- We will always attempt to populate `created_at` from a property, when Org file is loaded
- If it's not the right property, user can change it and re-populate `created_at` from a different one
- User might or might not store the created-at as a property when exporting to Org file

Let me know what you think.

I'll see to merge this now and modify the code in this direction, and we can continue from there in #114.

nevenz:
BTW, can you squash these and either rebase or open a new PR, or mind if I squash them? There's a lot of syncing with upstream and merging commits, and it's all from way back (my bad).

cmdd:
Sure, feel free to squash the commits! (I think Github has a merge-and-squash feature, if I'm not mistaken). All of your proposals make sense, although I am curious as to how this feature would interact with sorting all properties in general (maybe that's a question for a separate issue...)

In the meantime, once this gets merged I'll try to implement your suggestions in a new PR as soon as possible.

nevenz:
This (with bunch of refactoring) is in master now.

Let me know (in #114 or #254) if something is not working as you planned.

Thanks!

PR #135, opened 2017-07-31
Feat/fab sync

tulth:
I added an optional sync floating activity button in the bottom left.
Enable in the preferences under the sync section.
Let me know if there are conventions I should be following to fix to make it mergeable.

### DIFF
--- a/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
+++ b/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
@@ -507,4 +507,10 @@ public class AppPreferences {
                 context.getResources().getString(R.string.pref_key_sync_after),
                 context.getResources().getBoolean(R.bool.pref_default_value_sync_after));
     }
+
+    public static boolean floatingSync(Context context) {
+        return getDefaultSharedPreferences(context).getBoolean(
+                context.getResources().getString(R.string.pref_key_floating_sync),
+                context.getResources().getBoolean(R.bool.pref_default_value_floating_sync));
+    }
 }
--- a/app/src/main/java/com/orgzly/android/ui/MainActivity.java
+++ b/app/src/main/java/com/orgzly/android/ui/MainActivity.java
@@ -1494,8 +1494,9 @@ public class MainActivity extends CommonActivity
             ((DrawerFragment) drawerFragment).setActiveFragment(fragmentTag);
         }
 
-        /* Update floating action button. */
+        /* Update floating action buttons. */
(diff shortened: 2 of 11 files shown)


nevenz:
I don't think it's worth adding a new FAB just to save one swipe (to open a drawer) or click (to open the overflow in the top right corner)?

Also, we'd end up with 3 different ways to start a sync, all one click/swipe apart.

FAB on the left is kinda non-standard too, we'd want to move it above the current one. And I was planning to add the search there.

I think we should just focus on making sync start automatically when needed (your next PR).

tulth:
I agree that if the auto sync works well enough, the floating button is unneeded.

I still need to add autosync after a note is modified.  I will add that tonight maybe, but what is unfortunate is that right now, it will sync all notes when any note is modified.  

I would like to add a follow on feature to make the notes more autonomous so they can sync themselves to solve the above issue.

nevenz:
> I still need to add autosync after a note is modified. I will add that tonight maybe, but what is unfortunate is that right now, it will sync all notes when any note is modified.

> I would like to add a follow on feature to make the notes more autonomous so they can sync themselves to solve the above issue.

What do you have in mind? Since we need the complete file to be able to sync it (with Dropbox for example). And I don't think it's possible to modify the file without having to rewrite everything after modification (unless the size remains the same).

Per-note syncing would be possible with server support, which is something I do plan to implement (syncing to orgzly.com, like you can with Dropbox now, but more efficient due to this).

Maybe I'm misunderstanding what you mean by "syncing"? There has been some confusion there in other threads.

tulth:
I just mean sync on a note book level, not am everything level.  The terms are a bit ambiguous!

In other words, if i have todo.org and notes.org:
right now if i sync, both files are synced.  There isn't any per notebook granularity on the syncing that I see.  I would like to update this do that if I change todo.org, and I autosync after the change, it would only update the corresponding todo.org file, not the notes org file.

nevenz:
> In other words, if i have todo.org and notes.org:
right now if i sync, both files are synced. There isn't any per notebook granularity on the syncing that I see. I would like to update this do that if I change todo.org, and I autosync after the change, it would only update the corresponding todo.org file, not the notes org file.

This should be the case now. File is not created unless its notes are modified. Nor downloaded unless its modification time (or Dropbox revision if you're using that method) is changed.

tulth:
ok I will not make the per notebook sync change.  I will call the master sync and let its logic decide what should be updated.

nevenz:
I'll close this one and we can continue in #136, or better a new PR without FAB.

PR #136, opened 2017-07-31
Feat/sync on resume

tulth:
Includes previous sync floating activity button feature.
Adds optional feature runs sync whenever the app moves to the foreground.
Defaults disabled, check to enable in prefs.

### DIFF
--- a/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
+++ b/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
@@ -507,4 +507,10 @@ public class AppPreferences {
                 context.getResources().getString(R.string.pref_key_sync_after),
                 context.getResources().getBoolean(R.bool.pref_default_value_sync_after));
     }
+
+    public static boolean floatingSync(Context context) {
+        return getDefaultSharedPreferences(context).getBoolean(
+                context.getResources().getString(R.string.pref_key_floating_sync),
+                context.getResources().getBoolean(R.bool.pref_default_value_floating_sync));
+    }
 }
--- a/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
+++ b/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
@@ -513,4 +513,10 @@ public class AppPreferences {
                 context.getResources().getString(R.string.pref_key_floating_sync),
                 context.getResources().getBoolean(R.bool.pref_default_value_floating_sync));
     }
+
+    public static boolean onResumeSync(Context context) {
+        return getDefaultSharedPreferences(context).getBoolean(
+                context.getResources().getString(R.string.pref_key_on_resume_sync),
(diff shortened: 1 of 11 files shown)


aaronbieber:
I found my way here from #8 because I recently decided to give orgzly another try (it's been a year or so perhaps) and on the very first day I found myself in a conflicted state because I foolishly used the new notification "new note" button without syncing first, which is very inconvenient to do (why open the full app when I have the notification?) 

As someone who uses Dropbox for everything and edits his org GTD files on mobile and desktop constantly throughout the day, I really want a solution like this one. 

I'm not a great Android developer, but I'd love to help with this if I can. I will test development builds or anything else within my abilities here.

tulth:
Update:
I broke out the sync on resume so that it is independent of the floating button.

i have also added some preliminary support for auto syncing whenever notes are updated.

I am running this today, and it seems like it is catching all my changes.

I do have an outstanding bug where the notebook will be synced but yet the notebook still displays the need sync icon.  I verified that the file on disk is correct. 

Is it worth a pull request for anyone else to try?

Right now, my approach is just to populate all the handlers in the sync fragment to kick off the sync service.  I also had to update actionservice to catch the mark as done in the notifications

I tried to use a content observer, but right now there are two major problems with this approach:
* no granularity on the reported URIs for changes.  All changes seem to report the same URI.  There are many content changes that the external sync does not care about.
* changes cause many observed change updates.  One small change may actually kick off half a dozen observer change notifications.

With that said, I would like to move to the observer approach in the future.  What are dev/maintainer thoughts on:
* making onchanged URI notifications more granular, or at least creating an content://.../updated namespace that the various clients could poke when a syncable change is made. Eventually book and note subfolder granularity could be added.
* making all other URI notifications fall into some other URI namespace.  For example, updating mod time could map to content://.../internal namespace.
* rework the sync service so that it would wait some small amount of time (say 100ms) on observed change notifications so that bursts of notifications could be combined to a single sync.

aaronbieber:
I pulled down this patch, applied it, and have been running it with no problems.

The one thing that is "missing" for my use-case is syncing *before* adding a new item. With the notification-triggered "new note" functionality (which I love), you can get into a situation where the file was changed on Dropbox, then you add a new note on mobile without syncing, and voila, you have a conflict.

I was able to trigger the sync intent from the ShareService where that action is started, and it does work, but the notification never displays. I'd love a little help figuring that out.

I think splitting the FAB from the sync-on-resume is a good move.

aaronbieber:
One piece of feedback I have on the original MR is that the preference checkbox text is too long for a single line and is shortened with an ellipsis (...) in such a way that the option's meaning isn't clear. This is on my Google Pixel. I think the text needs to be wrapped, or the option rephrased. Probably wrapping is more compatible with even smaller devices.

tulth:
closing this PR.  Replaced by pull request #159

Issue #117, opened 2017-06-27
Note count per notebook


fresheed:
As discussed in #112, requested features should be implemented one by one so here I would like to clarify note count. 

- Where do you think it should be put - to widget and/or note view? As for me, counter at widget is most needed one. 
- Is corresponding setting needed for that or it should always be displayed?

nevenz:
Are you referring to #62 widget or something else?

I though it would be a count next to a notebook - in the list of notebooks and/or drawer. I don't think there needs to be a setting for that.

PR #137, opened 2017-07-31
Notes count in notebooks list

fresheed:
This is the feature discussed in #117 . I've implemented detail view which displays amount of notes (without taking into accout their hierarchy). To do this, I've added corresponding subquery to book view. Also some refactoring of BooksFragment was performed.
I'm not sure where test for this should be placed. For this moment it belongs to BooksTest.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/BooksTest.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/BooksTest.java
@@ -1,6 +1,7 @@
 package com.orgzly.android.espresso;
 
 import android.os.Environment;
+import android.os.SystemClock;
 import android.support.test.rule.ActivityTestRule;
 
 import com.orgzly.R;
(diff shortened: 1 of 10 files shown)


fresheed:
Rebased previous commits to latest master state

nevenz:
We probably want this added to Settings's "Notebook -> Details", like other details, so it can be toggled.

And if you can rename variables to use camel case it would be great.

Thanks! (And sorry for very late reply.)

nevenz:
Oh and one issue - cut notes and root note are also counted.

Check `DatabaseUtils.WHERE_EXISTING_NOTES`.

fresheed:
Done

nevenz:
Few changes I made:

* Removed one DB_VER
* Moved line after modification time, to keep last-action last (so it's easy to spot, as it's important) 
* Icon changed (the previous one was too strong compared to others)
* Extracted strings and used plurals (we had "Contains 1 notes")

Thanks again!

fresheed:
Thank you for great efforts on making org portable!

PR #159, opened 2017-08-25
Feat/smarter sync

tulth:
Looking for feedback.  Its not yet ready for merge.

Has two features:
* onResume sync
* sync after notes are updated

Fixes:
* sync on resume text wrapping issue

Removed:
* floating sync activity button from prior pull request

Bugs:
* (FIXED, see below) notes after sync still indicate they need sync.  Reporting problem or race condition.

missing:
* does not sync before applying changes for notes updated, would be nice to sync from file, then sync to file


tulth:
Well, I traced the bug to <file>.lastModified() truncates to seconds resolution, not milliseconds.  So basically its unix time always ends with 000.  When I autosync, it happens quickly after, and the truncated time shows the file being out of date.

aaronbieber:
The string "Start sync after new note is created" is still too long to be fully displayed on the Pixel. Can we rephrase to "Start sync on note creation" and "Start sync on note update"?

start_sync_line_length

aaronbieber:
Org Mode and Orgzly are likely to be used by people with above average technical skill or interest, but can we also rephrase "Sync onResume" so that it doesn't reference the Android event? Maybe just "Sync on resume"?

nevenz:
> Well, I traced the bug to file.lastModified() truncates to seconds resolution, not milliseconds. So basically its unix time always ends with 000. When I autosync, it happens quickly after, and the truncated time shows the file being out of date.

Nice find. But I think current fix might cause issues too. For example if user change is queued due to running sync (#10), it's then executed right after sync finished. Local modification time could be less then 1s after remote file's. And the change wouldn't be detected.

I'm not sure how to fix this. It seems it might be possible to get milliseconds in the future with newer Android and Java versions. If we end up using this fix, we should at list limit it to local repositories. Dropbox should't have this issue.

nevenz:
As for Settings - I was thinking to have only one entry ("Auto-sync") below "Repositories" which would then open a new subscreen.

There, we'd have one preference to toggle auto-sync, and others would depend on it. 

Big fat warning would be on the top, above all, explaining potential issues (blocked while syncing is in progress, long db export time, whole notebooks transferred over net every time etc.).

Might even throw "experimental" or "beta" next to "Auto-sync" (yes, this thing is freaking me out, it's why I've been postponing it for so long 😄 ).

tulth:
> For example if user change is queued due to running sync (#10), it's then executed right after sync finished.
I only use this for file-based syncs, with syncthing.  What are the other use cases that cause it to be slow?  Slow hardware?  Large files?  I don't use dropbox but I thought it would appear as just files.  I am unaware of a network option, are there people with mods to support scp or webdav?
I am asking because I do not understand this part of the problem.  I need to understand this better so this isn't a mess when more try it.

> I'm not sure how to fix this. It seems it might be possible to get milliseconds in the future with newer Android and Java versions. If we end up using this fix, we should at list limit it to local repositories. Dropbox should't have this issue.
I agree we should be able to get millis.  I can see it has them on the filesystem in a shell.  I would be interested if anyone else has this problem or if it is limited to my build setup.  I am using java 1.8

I was thinking we could use a hash of the file which could be checked if the mod time was in a window of uncertainty.

> Might even throw "experimental" or "beta" next to "Auto-sync" (yes, this thing is freaking me out, it's why I've been postponing it for so long 
I totally agree.  This is very much alpha.  There are probably many race conditions remaining.
I view this current work as roughing out the idea.  I think with better understanding there could be a much better design for sync.

nevenz:
Orgzly used to call `notifyChange` for different uris (what you mention in #136), but it was pretty messy. I don't remember exact issues, but loaders were not being triggered when needed, so I just switched to using a single uri. We might want to reconsider this, but it might be a lot of work which is not worth it.

`notifyDataChanged` in `Shelf` is related to `syncAfterNoteUpdated`. We should at least have both in the same class. I think `Shelf` might be a better place. Then some cleaning up could be done probably.

nevenz:
> I only use this for file-based syncs, with syncthing. What are the other use cases that cause it to be slow? Slow hardware? Large files? I don't use dropbox but I thought it would appear as just files. I am unaware of a network option, are there people with mods to support scp or webdav?
I am asking because I do not understand this part of the problem. I need to understand this better so this isn't a mess when more try it.

Getting modification time of remote file is done by repository implementation, so it will depend and could be done differently for WebDAV, SCP, etc.

For Dropbox, its API is used (`files.getServerModified().getTime()`). Not the downloaded file's `File#lastModified`. But for local directory (which is just one of the repository types), that all we have. Which doesn't mean Dropbox doesn't do the same on their end, I haven't checked the values returned.

> I agree we should be able to get millis. I can see it has them on the filesystem in a shell. I would be interested if anyone else has this problem or if it is limited to my build setup. I am using java 1.8

I see it too, it's just the implementation. There's https://stackoverflow.com/a/24806116 for example.

> I was thinking we could use a hash of the file which could be checked if the mod time was in a window of uncertainty.

That would require first generating file from DB (which is slow) to calculate the hash. Perhaps we don't have to compare anything, just keep the `isModified` flag and change it on any modification.  Not sure if that's possible, I might be missing some use cases.

tulth:
sync_flowchart

Some thoughts on making a reliable sync architecture.

tulth:
Things missing from diagram:
* determine where/when to create file/dropbox observers
* handle cases where links or files are missing or changed

I think the existing code could solve the second item as long as the sync gets all data atomically.

nevenz:
I think we should keep all the individual preferences too (sorry if I wasn't clear). Users might want to fine-tune when they want to start the sync. So:

* Auto-sync
  * New note
  * Note updated or deleted
  * App opened

Last 3 would [depend](https://developer.android.com/reference/android/preference/Preference.html#attr_android:dependency) on "Auto-sync".

BTW, new screen is missing the action bar. I'm not sure what needs to be done for that.

Can we move the graph to https://github.com/orgzly/documentation ?

And I think we should treat that sync architecture you suggest separately from this issue and perhaps create a new issue for it? I'd like to merge this one and make it available in the next version, so we get as much testing and feedback as possible.

As for modification time issue, can we just start sync with 1s delay for now?

tulth:
I added back the 3 checkboxes and did some cleanup.
I added a placeholder checkbox for future repo observer capability.
I deleted the graph.

I see you asked for some overall checkbox but is unclear to me how that logic would work.  The autosync capabilities are just the 3 (soon 4) check box features.  I probably do not understand what you are getting at.  Let me know in more detail and I'll implement.

> BTW, new screen is missing the action bar. I'm not sure what needs to be done for that.

I spent a number of hours on this and browsing stack overflow and I cannot seem to get any of the proposed fixes to match what we are doing and ultimately work.
Any ideas?

> As for modification time issue, can we just start sync with 1s delay for now?

I think you are talking about this change:
```
    public boolean isModifiedAfterLastSync() {
        // Android does not report last file modification time in milliseconds in all cases
        // The mod times are truncated to seconds in some/many cases.
        // so consider it synced if within the last second
        return lastSyncedToRook != null && modificationTime - lastSyncedToRook.getMtime() > 1000;
    }
```
I think instead of storing file lastModified for rook_mtime,  we store our own time stamp after the file is written.
Also, I created a utility for doing MD5 sums (not yet committed), and I am thinking we can test for file changes using this when the times are close.  We could store the MD5sum in the repo_version column for file/directory repos. 

I am considering this code segment in BookNamesake:
```
            if (book.getLastSyncedToRook().getRevision().equals(latestLinkedRook.getRevision())) {
                /* Revision did not change. */
                if (book.isModifiedAfterLastSync()) { // Local change.
                    status = BookSyncStatus.BOOK_WITH_LINK_LOCAL_MODIFIED;
                } else {
                    status = BookSyncStatus.NO_CHANGE;
                }
            } else { /* Remote book has been modified. */
                if (book.isModifiedAfterLastSync()) {
                    /* Uh oh. Both local and remote modified. */
                    status = BookSyncStatus.CONFLICT_BOTH_BOOK_AND_ROOK_MODIFIED;
                } else {
                    status = BookSyncStatus.BOOK_WITH_LINK_AND_ROOK_MODIFIED;
                }
            }

```
I think the goal here is to determine:
test1) book newer than the last repo sync
test2) repo newer than the last repo sync

if !test1 && !test2: nothing to do, already in sync
if test1 && !test2: repo newer, need to read repo and write to db
if test1 && test2: book newer, need to write book to repo
if test1 && test2: conflict

If we write our own now=DateTime() rook_mtime after the file is written, test1 still works, but it fixes the lack of time precision (assumes a book change does not slip in-between the file write and rook_mtime update).

If we use an md5sum-based rook_version for the test2 , we guarantee changes are picked up.  There is no issue with the truncated mtime making the repo seem older than the book.

I think without going and doing the big diagram work, I can implement these small changes and it will work a bit better than before.

tulth:
If the last 2 changes worry people, let me know.  I can put those on a diff feature.

tulth:
noted a problem with the last 2 changes.

Changing the DB over from mtime to hash and removing the 1000 ms window causes out of sync messages to show up on first install.  I'm going to use it tomorrow and see if anything else comes up.

nevenz:
> I see you asked for some overall checkbox but is unclear to me how that logic would work.

It would simply be used to enable or disable the other 3 preferences. It wouldn't have a value itself. No code for it or anything. It would basically be just another hoop to jump though to get to the other preferences.

Probably too much, now that I think about it again. A separate screen is good enough to grab enough attention.

> I spent a number of hours on this and browsing stack overflow and I cannot seem to get any of the proposed fixes to match what we are doing and ultimately work.

Yeah, I remember doing that too - I wanted to move reminders-related preferences to the separate screen.

I did plan to have settings in a separate activity (no drawer, potentially starting it from a widget in the future, and it just makes more sense, especially since repositories are in a separate activity). I think that might help. I'll take a look, we can leave it as it is for now.

> I think instead of storing file lastModified for rook_mtime, we store our own time stamp after the file is written.

What about `getBooks()` though? Not sure if we can have it return a different time.

> Also, I created a utility for doing MD5 sums

Let's move these two to a separate issue/PR and continue there. They both change the way things work now, which needs much more testing then a new (beta/experimental) auto-sync.

Unless you think 1s sync delay can cause other issues? Or is too much time? We do need at least a workaround around that for now.

tulth:
> Unless you think 1s sync delay can cause other issues? Or is too much time? We do need at least a workaround around that for now.

I will be honest, I did not understand what you were getting at until I gave it some thought.
I have not tried it but I think this will work the caveat that there are some race conditions and the sync lockout will be more noticeable for file/dir based repos.

I will try this tonight or tomorrow.

> Let's move these two to a separate issue/PR and continue there. They both change the way things work now, which needs much more testing then a new (beta/experimental) auto-sync.

ok, as above I will make this change soon and move these to another feature branch.

> What about getBooks() though? Not sure if we can have it return a different time.

I missed this; I will put the fix to this on the new feature branch.
I actually ran across this and have this fixed it my own testing branch that is basically both the snooze and smarter_sync features.

tulth:
Update
* I backed out the hash changes
* made it wait if it the truncated now (and by extension possibly file mtime) would be less than the book mtime

It seems to work pretty well as long as you do not try to update things while the sync is running.
The syncs can take longer now, and it is possible slip in changes while its syncing.
I am not sure of the exact mechanics but I was able to select a note for move and move it a bunch rapidly then ended up in a desynced state.

tulth:
I just added a new activity for the experimental auto sync settings screen so it renders correctly and has the settings style

nevenz:
> I just added a new activity for the experimental auto sync settings screen so it renders correctly and has the settings style

Cool.

Remaining issues I have:

- [ ] Don't start sync if repositories are not set (we keep getting a warning now)
- [ ] Moving notes is usually a multi-step process, sync when user is done (might be tricky, use actionModeDestroyed?)
- [ ] Promoting and demoting notes is ignored (could have the same issue as above)
- [ ] Move sync-starting to Shelf
  - so it's next to (related) notifyDataChanged for now
  - so it's not both in ActionService and SyncFragment
  - it's started when there is (possibly) no need (setStateToDone)
- [ ] syncAfterNoteCreate is checked after setStateToDone (should be syncAfterNoteUpdate)
- [ ] Delay sync-starting 1s, don't sleep before saving each book to repo
- [ ] isModifiedAfterLastSync is still checking that difference is > 1s

Last two are important, others only matter if auto-sync is enabled.

tulth:
- [ ] Delay sync-starting 1s, don't sleep before saving each book to repo
- [X] isModifiedAfterLastSync is still checking that difference is > 1s
- [ ] Don't start sync if repositories are not set (we keep getting a warning now)
- [ ] Moving notes is usually a multi-step process, sync when user is done (might be tricky, use actionModeDestroyed?)
- [ ] Promoting and demoting notes is ignored (could have the same issue as above)
- [ ] Move sync-starting to Shelf
  - so it's next to (related) notifyDataChanged for now
  - so it's not both in ActionService and SyncFragment
  - it's started when there is (possibly) no need (setStateToDone)
- [ ] syncAfterNoteCreate is checked after setStateToDone (should be syncAfterNoteUpdate)

tulth:
I moved the sync delay to the asynctask sync task (what a mouthful!).
I tried delaying in the service before the task launch and it blocks the UI thread.
It seems like it works ok in quick testing...in the future with the hash based-check and file locking, this problem will go away for good.

tulth:
Note, some cases remain where unneeded autosyncs:
- promote/demote/move request that result in no actual change
- schedule changes that give no change
- opening the note and making no changes and hitting the checkbox (Shelf::updateNote)

I looked into skipping no-change note updates, and it is not trivial.
There is no equality defined on Note, and there are a number of moving parts.

I think if we fix the above, any non trivial fix should move to another pull request.

Thoughts?

nevenz:
> I moved the sync delay to the asynctask sync task

I was thinking to sleep only if sync is started automatically. Now we'll delay user triggered sync too. This could be a flag in intent (like `EXTRA_AUTOMATIC`) set in `autoSync()`.

> Note, some cases remain where unneeded autosyncs:

Another minor one - in `paste()` when `batch` is null - i.e. when there is nothing to paste.

> ... paste() ...
>         createSync();  // should this be updateSync?

I think it should, as cut comes first. It's the same as if you moved the note.

> I looked into skipping no-change note updates, and it is not trivial.

There's `isNoteModified()` in `NoteFragment` already, which I think we can simply check before `onNoteUpdateRequest()`. I should have done that in the first place probably.

> I think if we fix the above, any non trivial fix should move to another pull request.

In general, I do prefer multiple smaller PR, they are easier to manage.

Anyway, I can merge this whenever, let me know.

tulth:
fixes:
* changed paste to use updateSync instead of createSync
* added the no-change checks you pointed out (paste, no update)

I would like to hold off on the delay change because I think the upcoming hash based file checking will solve this more cleanly - there will be no delay at all, so no need to discriminate user directed vs autosyncs.

I think it is ready for a merge test if you do.

tulth:
I think I should add resumeSync to the share activity.  I will add this tonight if there is no disagreement

nevenz:
I did remove that sleep when user starts the sync, it felt too weird waiting.

I got a few "Too many requests" errors from Dropbox, since I started using this (all options enabled). We'll need to limit the sync for that.

I often create multiple notes one by one, or create one then change its state etc. Dropbox errors aside, syncing between each action is too much and unnecessary. Of course I could turn off the option to sync after modification or creating new note, but I think then I'd be back to square one. Option to "Start sync when you *leave* the app" might be useful though? Some periodically scheduled sync could be thrown in as well.

Anyway, it's a good start, and we'll keep on improving it. I might bury it even deeper in settings though, at least in a next release.

Thanks!

tulth:
i have been considering rate limiting the syncs too.  Maybe this would solve these issues?

I cannot find any source on Dropbox limits.  Any suggestions?  I was thinking rate limit to 1 second for local files but maybe something more like 15 to 30s is appropriate for dropbox?

aaronbieber:
Could it be set up to run a sync after X changes or X time, whichever comes
first?

It feels like we need some kind of a queue within the sync service so that
many rapid sync events can be de-bounced (this was mentioned before) and so
that a queued request can run after some set time interval even if the app
isn't foreground anymore.

I am still using a fork off of tulth's original feature that triggers a
sync when the "new note" share activity is triggered and it's working great
for me, though the sync notification failed to display in Nougat. I just
upgraded to Oreo and now it does display. Weird Android bug perhaps?

nevenz:
Yeah some smarter logic for when to sync (or not to sync) is needed and should be in one place.

I don't think Dropbox needs much limiting, we just need to make sure not to ping it every second or two for start.

Issue #153, opened 2017-08-19
snooze option


tulth:
It would be nice to have a snooze button for reminder notifications. 

Also it would be great to see the snooze appear in Android Wear peripherals.

PR #160, opened 2017-08-28
Feat/snooze

tulth:
Features:
* Notification action support for android wearables.  Tested to work with pebble fw 4.3.
* snooze feature for reminder pop ups.
* setting for snooze time in preferences, defaults to 10 minutes.

Looking for feedback.
The lifecycle for the snooze was much more complex than I expected. See diagram:

snooze_diagram

If it can be simpler, let me know.


tulth:
address issue #153

tulth:
Just a comment:  I have been running with this feature all week and its working great!  

I use the snooze quite a bit when I am wrapping up one task but I do want a reminder on the next task.

app/src/main/java/com/orgzly/android/reminders/ReminderService.java
```diff
@@ -412,7 +483,29 @@ private PendingIntent markNoteAsDonePendingIntent(
...
+        intent.setAction(AppIntent.ACTION_REMINDER_SNOOZE_REQUEST);
+
+        intent.putExtra(ActionService.EXTRA_NOTE_ID, noteId);
+        intent.putExtra(ActionService.EXTRA_SNOOZE_TIMESTAMP, timestamp);
+        intent.putExtra(ActionService.EXTRA_NOTIFICATION_TAG, notificationTag);
+        intent.putExtra(ActionService.EXTRA_NOTIFICATION_ID, notificationId);
+
+        return PendingIntent.getService(
+                context,
+                Long.valueOf(2 * noteId + 1).intValue(),
```
nevenz: Did you have issues with this being just `noteId` and leaving the same above in `markNoteAsDonePendingIntent`?

This shouldn't need to be unique if intent is different (even by having only a different action).

tulth: you are right, if the actions are different, they will be considered different.  I will make them both note id, not 2*id+x.

nevenz:
We are currently scheduling snooze job relative to when reminder ran. I think we want job scheduled X minutes after hitting snooze button? 

For example now, if snooze time is set to 5min and you hit the button 4min after notification is displayed, it will be displayed again in 1min.

Or if you leave notification longer then snooze time, another one will popup 1s after you hit snooze.

licaon-kter:
> I think we want job scheduled X minutes **after** hitting snooze button

Yes, this.

tulth:
You are right, it maintains an exact relationship to the base alarm time.

I can make it a check box to snooze relative to press or alarm...
I use it and prefer it to be exact off the base time...

Say I have an upcoming meeting, i want a couple early reminders at exact times.  
Because I know the alarm time and snooze interval time, i know exactly what time it is when the next alarm goes off.

Or if someone needs to be up by 7:15 am, you could set an alarm a few intervals early and know how many intervals you have up to be up the 7:15 am time.  If you sleep thru a few minutes of the snooze times, the upcoming alarms drift off and are no longer tied back to any specific time.

It seems like if you don't care if the snooze time is accurate to any specific time, and you may not care if its 10 minutes after the button press vs 10 minutes from the alarm start.

tulth:
i will update so the snooze will always be a positive interval; no more immediate snooze.

nevenz:
> Say I have an upcoming meeting, i want a couple early reminders at exact times.

_Planning_ to hit snooze every time just doesn't sound right to me. But that's just me and possibly my misunderstanding of what the snooze is. :) And of course, we all have different systems/workflows.

To me, snooze is a simple "I'll deal with this in 5min" kind of thing.

For the case you mention we should look into improving the timestamps and specifying repeaters and when to trigger the reminders. Support for `<2017-09-01 7:00-7:15>` would go in that direction. Plain timestamps (so multiple can be used in one note) or warning period support too. Perhaps something else...

> I can make it a check box to snooze relative to press or alarm

Still, with above said, I'm fine with a preference (instead of checkbox) for this. Either to change the behaviour of the button, or add another. This is to keep the notification simple and so that it doesn't require too much thinking or clicking (checkbox + action). Would that be OK?

tulth:
i added the preference selection to start the snooze interval from the alarm time or button time.  I would like feedback on it - definitely with the the verbiage, and whatever else.

Also, would you like me to rebase it to master to keep things clean?  or just merge in master updates?

nevenz:
If you can rebase it, great, though it's not a big deal, I'll merge it.

One issue I'm having - when note has both scheduled and deadline times, wrong time can popup in notification after snooze. `onSnoozeTriggered` is just using the first time it gets for the note.

I have yet to try `alarm` and wrap my head around it. I just don't see that as "snooze". Changing the setting (in one direction or the other) after hitting snooze also comes to mind as something that needs testing, though I might be nitpicking too much there.

tulth:
I will fix the scheduled/deadline problem tonight or tomorrow night. 

I might have misunderstood on the preference, was the list preference with selection options what you had in mind?

nevenz:
> I might have misunderstood on the preference, was the list preference with selection options what you had in mind?

No, preference is fine. I was referring to *testing* the case when user hits snooze, then changes the preference before notification is displayed again. Just a random thing that occurred to me when I was thinking about testing `alarm` preference.

tulth:
Added tracking for scheduled vs deadline through the snooze lifecycle.

Only odd thing is that if deadline + schedule fires for the same note at the same time, the notification is lost.  I believe this is because of the combining of like intents - same action, same id.

tulth:
I am thinking of using one of three approaches:
* add separate actions for snooze scheduled vs snooze deadline.  
* do the 2*noteId + 1 vs 2*noteId interlace trick for the intentId
* Add some "data" URI to the intent to make them distinct - setData(Uri data).  I am less sure about this one.

Any suggestions?

tulth:
>> Only odd thing is that if deadline + schedule fires for the same note at the same time, the notification is lost.
> Which notification, the one that should be displayed after snooze? Or the original?

For deadline and scheduled firing at the same time, only one notification is produced.  I expected 2 notifications.
I believe they match intent equals so they get combined.  I think whichever arrives first is what it is labeled as in the notification.
I am not sure what we actually want to happen in this case.
I could leave it as is.
Or apply one of the above fixes.

nevenz:
> For deadline and scheduled firing at the same time, only one notification is produced. I expected 2 notifications.

From implementation - yes, but normally I'd want to see a single notification per note, describing that it has both deadline and scheduled set or whatever.

I guess it's fine for now. I think some things will need to be rewritten in the future.

tulth:
ok let me know if there is anything else to make this ready for the merge!

PR #167, opened 2017-09-04
Add preferences for tags-column & org-indent-mode

Sodel-the-Vociferous:
I added these features to the org-java back-end months ago. Sorry it took so long to create this patch for Orgzly, but here you go!


nevenz:
BTW, I moved these preferences to a separate screen, to be clear that indent mode and indentation per level are used *only* for tags indentation currently.

Thanks!

Sodel-the-Vociferous:
Good call. :) Thanks for merging!

Issue #149, opened 2017-08-14
Implement support for `(org-tags-view)` match syntax


wcerfgba:
The match syntax for `(org-tags-view)` is very flexible and allows for more advanced queries than the current search functionality. I appreciate the existing search syntax, and the ideas in [the extended search proposal], and I think that, although the `(org-tags-view)` syntax is more unwieldy (especially on mobile) than the reduced syntax already available, it would be good to support the 'official' syntax from desktop. Additionally, if this was implemented in a well-architected way, most of the components required to implement the current extended search proposal would also be built in the process.

From release 9.0.9 of Org I have found [the definition of `(org-tags-view)`](http://orgmode.org/cgit.cgi/org-mode.git/tree/lisp/org-agenda.el?id=release_9.0.9#n4811), and [the definition of `(org-make-tags-matcher)`](http://orgmode.org/cgit.cgi/org-mode.git/tree/lisp/org.el?id=release_9.0.9#n14495), which contains the parsing of the match query and the logic for building a matcher function. The code is a little over 100 lines of elisp, so I am interested in porting it to Java and implementing a new search function that supports this syntax.

Open to hearing any thoughts on this. Thanks for your time and for all your work on Orgzly so far, it's great! :grin:

nevenz:
I wasn't thinking about supporting that syntax (current one will eventually cover everything), but I was about converting queries between the two. So implementing this, apart for being useful to users who want to use Org mode's syntax, is a step toward that as well. So it's very welcome. 😃 

It would be great if the code could be reused (everything after syntax parsing I guess).

One minor thing to resolve - how/when to choose which syntax to use. Auto-recognition would be ideal, but I'm pretty sure it would be ambiguous. So either a switch next to a search, or an option in Settings I guess. As I don't think users would often (enough) mix the two, I'd vote for an option in Settings (it's easier to implement as well).

Thanks!

wcerfgba:
Hi @nevenz ,

I have some thoughts on architecture, then I will address your point about the UI.

I can see there is a `SearchQuery` class which exposes the domain language for a search query, and that this is used in `com.orgzly.android.provider.Provider#queryNotesSearchQueried`, where we construct the `SearchQuery` from a string, which is then used to build the SQL query.

[...]

Issue #124, opened 2017-07-10
Multiple state search


borablanca:
I am using multiple states in my workflow s.t. TODO -> NEXT -> DONE. So it will be useful if multiple state search can be added for similar users using this kind of multiple states. For example if I want to see both TODO and NEXT stated items, I may search by i.todo.next as an idea. Or using i multiple times such as i.todo i.next in the search query.

PR #169, opened 2017-09-09
"Or" Searches

Sodel-the-Vociferous:
This implements the "or" operator in searches.

Now, searches like `tn.tag1 b.book1 or tn.tag2` are possible.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/QueryFragmentTest.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/QueryFragmentTest.java
@@ -493,4 +493,21 @@ public class QueryFragmentTest extends OrgzlyTest {
         onListItem(0).onChildView(withId(R.id.item_head_title)).check(matches(allOf(withText("Note A  a"), isDisplayed())));
         onListItem(1).onChildView(withId(R.id.item_head_title)).check(matches(allOf(withText("Note D"), isDisplayed())));
     }
+
+    @Test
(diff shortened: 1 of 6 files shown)


Sodel-the-Vociferous:
Should provide a solution for #124.

IvanMalison:
Whats the status of this?

nevenz:
Ivan Malison <notifications@github.com> writes:

> Whats the status of this?

It's in my queue.

nevenz:
Thanks!

This is likely to be rewritten soon though, there has been some talk about it in #149. We probably shouldn't expose `currentGroup` like that, although I do realise re-writing tests could be tricky.

I also feel like we need to support grouping (and explicit `and` operator) together with it, to avoid questions, requests and possible confusion -- for example, first thing I did when I tried this is expecting `i.todo or i.next i.work` to be (`i.todo or i.next) and i.work`.

( I know I was suggesting `and` to have precedence in the past, but that was with *explicit* operator specified. Though I think the logic is still fine as it is now. )

Sodel-the-Vociferous:
Yup, I knew this was all likely to be rewritten. Otherwise I'd never expose currentGroup like that -- I agree it's a "code smell". 
(And, you're right that I kind of violate the principle of least astonishment with grouping. But until we require that either "and" or "or" be explicitly added between all query terms, one of them has to take precedence. And I definitely don't want to break backwards compatibility! Anyway, I'll continue in #149 if I get any decent ideas.)

PR #172, opened 2017-09-11
add hash-based file sync check

tulth:
Starting point for hash based file check/sync
During the changeover to hash file revisions, all files will flash as out of date.
This does not solve some of the other race conditions on checks vs applied changes.
I do not like the global context but I do not know a better approach to getting access to the content provider.

### DIFF
--- a/app/src/main/java/com/orgzly/android/App.java
+++ b/app/src/main/java/com/orgzly/android/App.java
@@ -1,14 +1,23 @@
 package com.orgzly.android;
 
 import android.app.Application;
+import android.content.Context;
 
 import com.evernote.android.job.JobManager;
 
 public class App extends Application {
+
+    private static Context context;
+
     @Override
     public void onCreate() {
         super.onCreate();
 
+        App.context = getApplicationContext();
         JobManager.create(this).addJobCreator(new AppJobCreator());
+    }
+
+    public static Context getAppContext() {
+        return App.context;
     }
 }
--- a/app/src/main/java/com/orgzly/android/repos/ContentRepo.java
+++ b/app/src/main/java/com/orgzly/android/repos/ContentRepo.java
@@ -7,10 +7,13 @@ import android.provider.DocumentsContract;
 import android.support.v4.provider.DocumentFile;
 import android.util.Log;
 
+import com.orgzly.android.App;
 import com.orgzly.BuildConfig;
 import com.orgzly.android.BookName;
 import com.orgzly.android.util.LogUtils;
 import com.orgzly.android.util.MiscUtils;
+import com.orgzly.android.util.HashUtils;
+import com.orgzly.android.provider.clients.CurrentRooksClient;
 
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -75,11 +78,19 @@ public class ContentRepo implements Repo {
                                 "file.getParentFile()", file.getParentFile().getUri());
                     }
 
+                    VersionedRook lastRook = CurrentRooksClient.get(App.getAppContext() , getUri().toString());
+                    String rev = getRevision(file);
+                    long mtime;
+                    if (lastRook == null || !rev.equals(lastRook.getRevision())) {
+                        mtime = System.currentTimeMillis();
+                    } else {
+                        mtime = lastRook.getMtime();
+                    }
                     result.add(new VersionedRook(
                             getUri(),
                             file.getUri(),
-                            String.valueOf(file.lastModified()),
-                            file.lastModified()
+                            rev,
+                            mtime
                     ));
                 }
             }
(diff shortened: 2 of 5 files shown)


nevenz:
Check if 11ccbb8af323c2cb44cad74f57665fdbc4e2f541 fixes the issue that prompted this PR (I think we mentioned this fix).

Beside issues you mention, I see few more in this PR -- use of `current_versioned_rooks` (which I was planning to remove, as it's not used for anything), getting mtime from that table then storing the same value back, coupling of mtime and revision, revision losing a property of being unique over time which might cause issues in the future.

tulth:
I feel pretty strongly that hashing is required to maintain bulletproof file integrity, so I really want to make this happen.
Some example problems that can happen without hash: 
- user reverts file intentionally on external system and it shows older so the change is not picked up
- external system has poor mtime resolution
- times not being in sync across platforms

What steps should I take to make the pr acceptable, for example I could:
- search/walk the versioned rooks instead of using current_versioned_rooks
- add a hash column to the rook table if you want to keep rook_version monotonically increasing
- add context to version request calls or add the most current db rook to the call
- I will look into the read mtime being written back

nevenz:
I'd be fine with switching to using file's content hash as a revision for file-based repos, if the switch wouldn't cause conflicts for people who have local changes (issue you mentioned) and if this PR does *only* that (it it wouldn't modifiy remote book's mtimes which are also visible to the user).

But I don't think it's worth the trouble.

> user reverts file intentionally on external system and it shows older so the change is not picked up

It would be, first check in `updateStatus ` is for revision *change* and this qualifies.

> external system has poor mtime resolution

This should have been fixed, if you're referring to the issue auto-sync triggered. I'd be interested in any more issues that can be reproduced though.

> times not being in sync across platforms

We're only talking about local file-based repos, no?

In any case, I suggest we wait to see what will git sync need, as interfaces and checks might change or be moved around.

tulth:
I will rebase this to latest beta and try to scope this to only hash; I will see what I can do regarding the timestamp revision to hash change over.

I have been running this for a while now and its working great.

The motivation remains to make the syncing as reliable as possible.   

I think all modern systems that track changes have moved away from timestamps. In my specific case local files are shared with syncthing and I am not 100 percent confident in file timestamps working in all cases across all platforms.

hyiltiz:
Hash based syncs sound horrible; the git way of only keeping track of changes, then keeping a hash for the *history* is much more reasonable. 

Disclaimer: I started using the app for less than 3 days, with no idea how it works internally. Maybe syncing already works based on changes in file org file.

gideonite:
Is this issue still alive? Were these conflicts resolved somewhere somehow?

I'm a recent orgzly user. Everything is great! ...except for one thing which is that I find the sync functionality a bit cumbersome. All my files are synced whenever I hit sync even when there's no file changes. More importantly, any kind of trivial edits to the same file results in a conflict. Is there a better way?

nevenz:
@gideonite Yes, detecting changes still needs improvement. Main problem is how to compare data in DB with the (potentially remote) file. Comparing hashes of files' contents is ideal, but it requires DB export and downloading the remote file, both being slow.

This didn't get much attention mainly because I haven't seen any issues it would fix. But what you're experiencing does sound serious.

What's your setup (remote repository in turn synced someplace or...) ?

nevenz:
> I suppose a fair amount depends on the Dropbox API and what sorts of operations are supported. Namely, can you check the time stamp of the last change of a file?

Yes, this [is possible](https://dropbox.github.io/dropbox-sdk-java/api-docs/v2.1.x/com/dropbox/core/v2/files/FileMetadata.html#getServerModified--) and Orgzly uses it. That's why I'm confused you're seeing conflicts. Those should happen only if both Orgzly notebook and Dropbox file it's linked to are modified at the same time, before syncing.

crdx:
I'm also seeing strange sync behaviour. Sometimes after I edit a notebook and sync it afterwards it syncs but the sync icon stays in the menu in and in the list of notebooks. If I force load/force save it appears to do the action, but makes no change to the sync state either. Not sure if it's necessarily related to the same issue.

nevenz:
> I'm also seeing strange sync behaviour.

Thanks, I'll take a look. Also Dropbox?

crdx:
> Thanks, I'll take a look. Also Dropbox?

Yeah. I can't reproduce it reliably. It only happens on some notebooks, sometimes. But when it happens the only workaround currently is to delete the notebook and resync. And then that only helps until the next time.

I've checked within Dropbox's interface in those cases to see what the last modified time shows as, and it shows as you'd expect. If I hit sync with the last modified time being at least minutes ago (to be safe) it would still continue to say it needs syncing afterwards.

Any other info I can provide to help, let me know.

nevenz:
> I can't reproduce it reliably. It only happens on some notebooks, sometimes. But when it happens the only workaround currently is to delete the notebook and resync. And then that only helps until the next time.

And it's not a conflict or a missing link? One of the failed syncs on http://www.orgzly.com/help under "Status Messages" perhaps?

I've [made](https://github.com/orgzly/orgzly-android/commit/887194606c9ec5a88dbb8d7fbe31d815ce646d38) cases like that [clearer](https://github.com/orgzly/orgzly-android/blob/887194606c9ec5a88dbb8d7fbe31d815ce646d38/app/src/main/res/drawable-xxxhdpi/ic_sync_problem_black_18dp.png), as there was a same report about lingering sync icon which turned out to be failed sync.

PR #213, opened 2017-11-12
Guess at the changes needed for not wrapping content

neiljerram:
Counter-intuitively, it appears that the Android "wrap_content" is what
you want when you *don't* want the content to be wrapped.  (I think
because it means that the size of the widget is determined by wrapping
around the unwrapped content.)


neiljerram:
I'm closing this PR because I now take a different approach to viewing (on my phone) the tabular information that I was primarily interested in here.  (Specifically, I publish it to my web site.)

Issue #143, opened 2017-08-07
Quick actions on icon long press


EvanRain:
Like in Google Keep, Todoist and many other apps. Would be handy.

nevenz:
Assuming [App Shortcuts](https://developer.android.com/guide/topics/ui/shortcuts.html).

Danesprite:
I'd like to add two app shortcuts: one for creating a new note and one for syncing. So I guess they would be alternatives to the persistent notification you can turn on in the settings. Would these two be good? Any others anyone can think of?

PR #237, opened 2017-11-28
Add app shortcuts

Danesprite:
I've implemented two app shortcuts (#143): one for syncing and one for creating a new note.
App shortcuts apparently require activity intents, so I've implemented a 'trampoline' activity for the sync shortcut as explained in the docs [here](https://developer.android.com/reference/android/content/pm/ShortcutManager.html) under **Shortcut intents**. 

The new note shortcut works in exactly the same way as the new note notification.

I've done some manual testing using the Nexus 5X API 25 and 26 emulators and I'd be happy to write some automated tests, just thought I'd put this up first in case anyone has any suggestions.


nevenz:
Thanks! Looks good to me. These could probably be static shortcuts (is "trampoline" activity even needed for dynamic ones?) to keep it simple and be clear there is no data involved, though I'm fine either way.

Danesprite:
Hey Neven. Yeah, static shortcuts are much simpler. I was having some back stack issues with `ShareActivity` when I experimented with them before, but they seem to work properly now. 

As for the trampoline activity, it looks like it's required either way unfortunately. The app shortcut API works the same way for static and dynamic shortcuts in that if you give a non-activity target class or intent, when you try to use the shortcut a toast will pop up saying "App isn't installed." :disappointed:. I think the docs should be clearer on this really, but I guess shortcuts are still relatively new.

Just about to push some commits for using static shortcuts instead.

nevenz:
> The app shortcut API works the same way for static and dynamic shortcuts in that if you give a non-activity target class or intent, when you try to use the shortcut a toast will pop up saying "App isn't installed."

Ah, I see. Toast is a good idea anyway too, to get a feedback on click. Thanks again!

PR #248, opened 2017-12-07
Create Github issue template

licaon-kter:


nevenz:
Thanks, but there are currently only 6 open bugs from 120 total issues. There's many more features for this kind of template.

PR #256, opened 2017-12-19
Implemented a Snackbar when trying to mark a Note as Done 

Endi1:
Since the user can't mark a Note as Done, I implemented as Snackbar as discussed in #253 to warn the user that it can't be done.


nevenz:
I've moved this to `Shelf`, so that DB is not queried on the main thread. Also, Snackbar from `CommonActivity` is used, to be dismissed on user action, like others.

Thanks!

PR #260, opened 2018-01-01
Preset book on new note view from filter query

invadersmustdie:
This change implements a convenience feature by selecting the book from the filter defined in current widget view. As filter can get very complex, I've tried to narrow it down to first positive occurrence of a book. While this implementation is far from perfect IMHO this reduces a minor inconvenience and saves a couple of seconds a day.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/ShareActivityTest.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/ShareActivityTest.java
@@ -4,6 +4,7 @@ import android.content.Intent;
 import android.support.test.rule.ActivityTestRule;
 
 import com.orgzly.R;
+import com.orgzly.android.AppIntent;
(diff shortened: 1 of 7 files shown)


nevenz:
Thanks! Few issues:

* Crash when there is no filter selected (when widget is first added)
* Crash when query contains `b` searching for non-existent notebook
* When query doesn't contain `b`, default notebook (Inbox) is not selected (seems the first notebook is selected instead?)
* `foo or b.bar` doesn't find a book (again, first notebook is selected)

invadersmustdie:
Thanks for the feedback. I've fixed all the issues but I'm a bit concerned of the `NoteFragment.getInstance` behaviour. Per default `0` is passed, I've tried to make it backward compatible, but this strikes me as unclear area.

invadersmustdie:
Ups, must have missed that. Updated.

Issue #76, opened 2017-05-11
Support for plain timestamps


lahertel:
I was wondering whether there is any particular reason why only deadlines and schedules are implemented in orgzly, but not plain timestamps. I find the latter really helpful for appointsments etc. Also check [gnu.org](https://www.gnu.org/software/emacs/manual/html_node/org/Deadlines-and-scheduling.html#Deadlines-and-scheduling)

tkamat:
I will try to implement this, seems relatively straightforward despite my meager android knowledge.

nevenz:
> I will try to implement this, seems relatively straightforward

Thanks!

Just note there can be *multiple timestamps per note* anywhere in the content, which would need to be parsed and stored in a new table and linked to `notes`. Which also answers original question - that's the reason why this hasn't been implemented right from the start - it's more complex then scheduled and deadline times.

tkamat:
Yeah you are right, I have been having more trouble with this than I initially expected. 

One question, do you want the timestamps in the ui to show up only when it senses one present, or do you want an "add timestamp" button at all times?

nevenz:
> One question, do you want the timestamps in the ui to show up only when it senses one present, or do you want an "add timestamp" button at all times?

So considering that:
- Plain timestamps can be scattered throughout note's content
- There's no single place in note where one timestamp should be put when writing an Org file (like there is for scheduled and deadline times)

I think there are two parts of implementing full support for plain timestamps:

1. Parse note's content (for the first time and on after any content change) and search for timestamps. Store every found timestamp to DB (to a new table which would contain `note_id`).
2. Add a toolbar above note's content in note view (similar to the one for comments [on GitHub](https://help.github.com/articles/about-writing-and-formatting-on-github/#text-formatting-toolbar)). This toolbar will be used for many other things in the future (bold, italic, lists, tables, links, ...). Add "Insert timestamp" to the toolbar, which would open a dialog with date/time picker (already exists in Orgzly) and insert it to the note's content.

First one is I think much more important and will allow searching, showing notes in agenda, reminders, etc. And UI part would be pretty convenient (well, more then convenient, since typing org timestamp is a PITA).

Anyway, that was my plan and how I'd go for it. Perhaps there's a better one, or we can support only one plain timestamp per note, though I don't think that's a good idea.

It's a decent amount of work, but can be done step at the time. And it's not really complicated as it might sound. First part is probably enough to consider plain timestamps supported (at least I would).

[...]

PR #277, opened 2018-01-19
Support plain timestamps

fmdkdd:
This is a work in progress for fixing #76.

I've added a new relationship `note_content_times` and modified `Provider.loadBookFromReader` to populate the table.

I guess the next step is to update this table whenever the content changes, and to cleanup when the note is destroyed.

Then, we should be able to lookup in this table for agenda searches and reminders.

Am I on the right track?  Don't be afraid to tell me if not. :)

### DIFF
--- a/app/src/main/java/com/orgzly/android/provider/Database.java
+++ b/app/src/main/java/com/orgzly/android/provider/Database.java
@@ -16,6 +16,7 @@ import com.orgzly.android.provider.models.DbCurrentVersionedRook;
 import com.orgzly.android.provider.models.DbDbRepo;
 import com.orgzly.android.provider.models.DbNote;
 import com.orgzly.android.provider.models.DbNoteAncestor;
+import com.orgzly.android.provider.models.DbNoteContentTime;
 import com.orgzly.android.provider.models.DbNoteProperty;
 import com.orgzly.android.provider.models.DbOrgRange;
 import com.orgzly.android.provider.models.DbOrgTimestamp;
@@ -116,6 +117,7 @@ public class Database extends SQLiteOpenHelper {
         for (String sql : DbCurrentVersionedRook.CREATE_SQL) db.execSQL(sql);
         for (String sql : DbDbRepo.CREATE_SQL) db.execSQL(sql);
         for (String sql : DbNoteProperty.CREATE_SQL) db.execSQL(sql);
+        for (String sql : DbNoteContentTime.CREATE_SQL) db.execSQL(sql);
         for (String sql : DbPropertyName.CREATE_SQL) db.execSQL(sql);
         for (String sql : DbPropertyValue.CREATE_SQL) db.execSQL(sql);
         for (String sql : DbProperty.CREATE_SQL) db.execSQL(sql);
@@ -140,6 +142,7 @@ public class Database extends SQLiteOpenHelper {
         db.execSQL(DbCurrentVersionedRook.DROP_SQL);
         db.execSQL(DbDbRepo.DROP_SQL);
         db.execSQL(DbNoteProperty.DROP_SQL);
+        db.execSQL(DbNoteContentTime.DROP_SQL);
         db.execSQL(DbPropertyName.DROP_SQL);
         db.execSQL(DbPropertyValue.DROP_SQL);
         db.execSQL(DbProperty.DROP_SQL);
--- a/app/src/main/java/com/orgzly/android/provider/Provider.java
+++ b/app/src/main/java/com/orgzly/android/provider/Provider.java
@@ -35,6 +35,7 @@ import com.orgzly.android.provider.models.DbCurrentVersionedRook;
 import com.orgzly.android.provider.models.DbDbRepo;
 import com.orgzly.android.provider.models.DbNote;
 import com.orgzly.android.provider.models.DbNoteAncestor;
+import com.orgzly.android.provider.models.DbNoteContentTime;
 import com.orgzly.android.provider.models.DbNoteProperty;
 import com.orgzly.android.provider.models.DbOrgRange;
 import com.orgzly.android.provider.models.DbOrgTimestamp;
@@ -1403,6 +1404,12 @@ public class Provider extends ContentProvider {
                                 long propertyId = DbProperty.getOrInsert(db, nameId, valueId);
 
                                 DbNoteProperty.getOrInsert(db, noteId, pos++, propertyId);
+                            }
+
+                            /* Insert note's content times. */
+                            for (OrgRange time: node.getHead().getContent().getTimestamps()) {
+                                long rangeId = getOrInsertOrgRange(db, time);
+                                DbNoteContentTime.getOrInsert(db, noteId, rangeId);
                             }
 
                             /*
--- a/app/src/main/java/com/orgzly/android/provider/clients/NotesClient.java
+++ b/app/src/main/java/com/orgzly/android/provider/clients/NotesClient.java
@@ -140,8 +140,8 @@ public class NotesClient {
         }
 
         if (head.hasContent()) {
-            values.put(ProviderContract.Notes.UpdateParam.CONTENT, head.getContent());
-            values.put(ProviderContract.Notes.UpdateParam.CONTENT_LINE_COUNT, MiscUtils.lineCount(head.getContent()));
+            values.put(ProviderContract.Notes.UpdateParam.CONTENT, head.getContentString());
+            values.put(ProviderContract.Notes.UpdateParam.CONTENT_LINE_COUNT, MiscUtils.lineCount(head.getContentString()));
         } else {
             values.putNull(ProviderContract.Notes.UpdateParam.CONTENT);
             values.put(ProviderContract.Notes.UpdateParam.CONTENT_LINE_COUNT, 0);
--- a/app/src/main/java/com/orgzly/android/provider/models/DbNote.java
+++ b/app/src/main/java/com/orgzly/android/provider/models/DbNote.java
@@ -106,8 +106,8 @@ public class DbNote implements DbNoteColumns, BaseColumns {
         }
 
         if (head.hasContent()) {
-            values.put(CONTENT, head.getContent());
-            values.put(CONTENT_LINE_COUNT, MiscUtils.lineCount(head.getContent()));
+            values.put(CONTENT, head.getContentString());
+            values.put(CONTENT_LINE_COUNT, MiscUtils.lineCount(head.getContentString()));
         }
     }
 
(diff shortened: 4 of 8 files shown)


nevenz:
> Am I on the right track?

Looks good to me. `DatabaseMigration` will need to be updated BTW.

nevenz:
Do you want this merged so that orgzly/org-java#7 can be merged too and master branches stay in sync? And then open a new PR for the rest of the work?

Or you prefer doing all the work in this PR?

fmdkdd:
I think I' rather have the feature working completely before merging it into master.  Unless you want to use orgzly/org-java#7 for something else.  That one should be nearly complete.

I'm having trouble making the plain timestamps appear in the Agenda view.  I gathered that a query containing `ad.days` is handled by `AgendaFragment`, and that one currently collects the scheduled and deadline timestamps using `DbNoteView`.  So, should I extend `DbNoteView` to also include any active timestamp?  Or should I go another route?

nevenz:
> I think I' rather have the feature working completely before merging it into master. Unless you want to use orgzly/org-java#7 for something else. 

I only thought because by default, Orgzly uses org-java from `../org-java` (this is done in Orgzly's `settings.gradle`), so it might confuse some developers if they have both directories and Orgzly fails to compile after pulling new org-java code. I might just disable that lookup by default.

> I'm having trouble making the plain timestamps appear in the Agenda view.

Keep [WHERE_NOTES_WITH_TIMES](https://github.com/orgzly/orgzly-android/blob/master/app/src/main/java/com/orgzly/android/provider/Provider.java#L326) in mind too.

> I gathered that a query containing ad.days is handled by AgendaFragment, and that one currently collects the scheduled and deadline timestamps using DbNoteView. So, should I extend DbNoteView to also include any active timestamp? Or should I go another route?

That view is one-to-one, I'd rather not touch it for this. I think everything should be done during query time. I'll have a closer look where I think would be a good place to do this.

fmdkdd:
> I only thought because by default, Orgzly uses org-java from

Yeah I realized that.  It's actually convenient.  But we can wait for this PR before merging the `org-java` PR, right?

> That view is one-to-one, I'd rather not touch it for this. 

Ok good. I will try to update and cleanup the `DbNoteContentTime` then.

nevenz:
> But we can wait for this PR before merging the org-java PR, right?

I kinda want it merged to sort orgzly/org-java#4 out as it will be built on top of `OrgContent`.

But it's fine, no hurry with this PR.

In fact, master branches of two projects not being in sync could become a common scenario. I'll see to move things from `settings.gradle` to a file which won't be under VC, so developers can change it to use whatever they want, without having to worry about it when doing commits (I'm not sure if this makes sense, I have to first try it first 😄 ).

nevenz:
Umm, this is not merged. I did push changes with this included by accident, but immediately forced-pushed what I intended. GitHub apparently doesn't detect that.

BTW, orgzly/org-java#7 doesn't actually break anything. I don't know why I thought that. `content` was and will remain private and everything goes through `OrgHead`.

nevenz:
> BTW, orgzly/org-java#7 doesn't actually break anything. I don't know why I thought that. content was and will remain private and everything goes through OrgHead.

Scratch that, I was experimenting with Gradle configuration.

But perhaps we should make `getContent` still return a `String` and hide the changes from org-java user?

fmdkdd:
> But perhaps we should make getContent still return a String and hide the changes from org-java user?

I can do that.  Do we still want to expose the `OrgContent` object in `OrgHead` via a getter?

fmdkdd:
> Do we still want to expose the OrgContent object in OrgHead via a getter?

Well, we do actually use a getter for accessing the timestamps.  I guess we could expose a getter for these timestamps directly in `OrgHead`, but for now I've just named the getter for `OrgContent` `getContentObject`.

> Umm, this is not merged

That surprised me as well.  I guess I can keep pushing to this branch while it's WIP then, or should I open another PR?

nevenz:
>> Do we still want to expose the OrgContent object in OrgHead via a getter?

> Well, we do actually use a getter for accessing the timestamps. I guess we could expose a getter for these timestamps directly in OrgHead, but for now I've just named the getter for OrgContent getContentObject.

Cool. We'll see, we're likely to start adding more stuff to `OrgContent` which will be there just for convenience/speed (pointers to drawers and such), so if we can keep it hidden, great.

> That surprised me as well. I guess I can keep pushing to this branch while it's WIP then, or should I open another PR?

This branch and PR work for me. I don't use GitHub UI for merging. But as you wish.

fmdkdd:
Ok, I've pushed commits that should take care of updating the rows in `DbNoteContentTime` when a note is updated or deleted in `NotesClient`.  I've based the code off `DbNoteProperty`; let me know if any changes are required and if I've missed anything.

I'm not sure it's quite correct, as we do not check whether the rows in `DbOrgRange` are still used after a content change/deletion.  But, as I understand it, the issue also arises for note properties, so I've followed the existing behavior.

I've also updated `DbTimeView` in a separate commit, but haven't added tests for that one yet.

nevenz:
Is `OrgHead` missing `getTimestamps()`? I can't compile Orgzly with org-java at `7291948`. Or I'm doing something wrong?

fmdkdd:
Ah, sorry, it seems I forgot to push https://github.com/fmdkdd/org-java/commit/813e022ca5e1d1dda8b43824c6296df3cfab9e21 in time.

nevenz:
>  I forgot to push fmdkdd/org-java@813e022 in time

Thanks, merged now.

> I'm not sure it's quite correct, as we do not check whether the rows in DbOrgRange are still used after a content change/deletion. But, as I understand it, the issue also arises for note properties, so I've followed the existing behavior.

Yeah, there are few more places where cleanup is needed.

> Ok, I've pushed commits that should take care of updating the rows in DbNoteContentTime when a note is updated or deleted in NotesClient. I've based the code off DbNoteProperty; let me know if any changes are required and if I've missed anything.

Good way to check where timestamps need updating or deleting is following `Shelf#notifyDataChanged` and there that method is called.

fmdkdd:
> Yeah, there are few more places where cleanup is needed.

Do you mean, for this PR specifically, or for other entities in the DB?  If I create a note with a header containing properties, these are added to the `DbNoteProperty` table.  If I then delete the note with `NotesClient.delete`, the note is deleted from the `DbNote` table, but the row in `NoteProperty` remains.

Here is a test I added to `ProviderTest` in order to check that:
```java
@Test
public void testNoteProperties() throws IOException {
    shelfTestUtils.setupBook("notebook", "* Note\n:PROPERTIES:\n:DATE: 2011-09-11\n:END:");

    Note note = NotesClient.getNote(context, 1);
    List<OrgProperty> properties = NotesClient.getNoteProperties(context, 1);
    assertEquals(1, properties.size());

    /* Delete content */
    NotesClient.delete(context, new long[]{1L});

    /* The property should not be in the base anymore? */
    Cursor cursor = context.getContentResolver()
        .query(ProviderContract.NoteProperties.ContentUri.notesProperties(), null, null, null, null);
    assertEquals(0, cursor.getCount());

    cursor.close();
}
```

(I added a case for `NOTES_PROPERTIES` in `Provider.query` in order to check the rows, but it's only for testing this out)

Should these rows stay in the DB after delete?  To me, it seems they shouldn't, since care was taken to remove them in `NotesClient.update` for instance.  But maybe they are deleted somewhere else?  

And more to the point: Should I follow the existing behavior of properties for content timestamps, or should I make sure we delete them properly?  And is there a single place where notes are deleted?  I understand `NotesClient` is a helper for the operations in `Provider`, so if we handle the deletion of content times in `NotesClient`, some code that calls `Provider.delete` directly may bypass the necessary cleanup.

In fact, it might be that this cleanup is best handled by the DB itself, with foreign keys and delete on cascade.  What do you think?

nevenz:
> Do you mean, for this PR specifically, or for other entities in the DB?

I meant in general.

> Should these rows stay in the DB after delete? To me, it seems they shouldn't, since care was taken to remove them in NotesClient.update for instance. But maybe they are deleted somewhere else?

I agree they shouldn't stay, but I wasn't taking care of all the cases.

> And more to the point: Should I follow the existing behavior of properties for content timestamps, or should I make sure we delete them properly?

If it's easy and you find a good place to do it, great, but don't worry about it too much.

> In fact, it might be that this cleanup is best handled by the DB itself, with foreign keys and delete on cascade.

Yeah I agree.

I'd be generally more nitpicking above - the reason I'm not is because I'll be looking at [Room](https://developer.android.com/training/data-storage/room/index.html) soon, removing `ContentProvider` (or at least stripping it down just for the widget, if necessary) and probably changing DB schema a lot in the process.

I'll open a new issue or write something on Wiki about the plans, before I actually do anything.

fmdkdd:
> I'd be generally more nitpicking above - the reason I'm not is because I'll be looking at Room soon, removing ContentProvider (or at least stripping it down just for the widget, if necessary) and probably changing DB schema a lot in the process.
> I'll open a new issue or write something on Wiki about the plans, before I actually do anything.

Ok, good; just making sure we are on the same wavelength here. 👌 

I'll try to see if I can update the DB migration and look at `notifyDataChanged` calls to look for any egregious leak.

fmdkdd:
I've pushed to the branch.  I'm happy to report that I've been able to make the AgendaFragment list notes with plain (active) timestamps, by modifying WHERE_NOTES_WITH_TIMES and teaching AgendaFragment about content times.

Although, I'm not sure it's the right solution, as we are now running a second query inside `AgendaFragment.onLoadFinished`?  Otherwise we might be able to use a special purpose view in `Provider.runUserQuery` when agenda days are specified.  Let me know how you feel about this solution.

Otherwise, I added content times in `NotesClient.create`, and took care of them in `NotesClient.delete`.  Although, there are 3 different delete methods in there, so I definitely did not get them all.

I've yet to write the DB migration.

nevenz:
> modifying WHERE_NOTES_WITH_TIMES

Couldn't we simply use something like`_id in (select note_in from times_view)` here? I'm not sure why I didn't do that in the first place. Perhaps speed could be an issue, as that view is doing a lot. It would need to be tested and compared.

> Although, I'm not sure it's the right solution, as we are now running a second query inside AgendaFragment.onLoadFinished?

Yeah, something else needs to be done, tons of queries are fired now. And there should be none actually, as it's the main thread. You can feel quite a difference in speed when you do "ad.3", comparing latest commit with for example 7f5ecb06.

> Otherwise we might be able to use a special purpose view in Provider.runUserQuery when agenda days are specified. Let me know how you feel about this solution.

Even when "ad" is not specified, we'll want to be able to search by plain timestamps.

But that aside, I'd probably start by moving the whole merging business to Provider and doing at most one more query to get all the times. Or a single query with join.

fmdkdd:
Good points all around, I'll try to move that query to Provider, and re-use `times_view` if I can.

PR #316, opened 2018-04-18
Show sync action in app bar if there's room

mndrix:
Syncing is the most common action that I perform from the app bar.  If there's space, show Sync in the app bar instead of the overflow menu.

### DIFF
--- a/app/src/main/res/menu/activity_actions.xml
+++ b/app/src/main/res/menu/activity_actions.xml
@@ -25,7 +25,8 @@
 
     <item
         android:id="@+id/activity_action_sync"
-        app:showAsAction="never"
+        android:icon="?attr/ic_sync_24dp"
+        app:showAsAction="ifRoom"
         android:orderInCategory="10"
         android:title="@string/sync"/>
 


nevenz:
I was actually thinking about removing Sync (and Settings for that matter) from the menu, as they are both in the drawer.

Sync in particular is just swipe & click away (as it's always on the bottom of the drawer).

More importantly, the appearance of the button is not consistent now. For example (on Pixel 2 emulator), it's shown in notebooks fragment, but not in notebook etc.

Do you use Dropbox? Would enabling auto-sync for it (after some improvements to the logic) work for you? Syncing should become transparent in the future.

Thanks!

mndrix:
> Do you use Dropbox?

I use local sync.  I have a script in Termux which monitors the file system for changes and then triggers a git-based sync script.

> Would enabling auto-sync for it (after some improvements to the logic) work for you?

I've experimented with auto-sync, but for now I prefer manual sync.  Manual sync lets me batch my changes into logical groups which my sync script translates into a meaningful git commit.

> I was actually thinking about removing Sync (and Settings for that matter) from the menu, as they are both in the drawer.

That's understandable. If manual sync is an unpopular workflow, it makes sense for it to be available in only one place.  Would you like me to submit a separate pull request which removes Sync and Settings from the menu?  I'm trying to familiarize myself with the code and that'd be a good one.

nevenz:
> Would you like me to submit a separate pull request which removes Sync and Settings from the menu? I'm trying to familiarize myself with the code and that'd be a good one.

Sure, great, go ahead.

Issue #95, opened 2017-05-28
Refiling UI


dk87:
Hi,

it would be great if we can move an item between notepads :smile:

Best
dk87

nevenz:
You should already be able to cut the notes and paste them to a different notebook?

dk87:
could it, however, be useful to provide an interface (e.g. drop-down menu) in the detail view of a note providing this?

jonhind:
Yes please !!!

Somantiq:
Currently I view the cut-and-paste as a dangerous operation. If I am interrupted in the middle, I lose the original note. This proposal would solve it. Or, maybe, make the "cut" operation tentative like in Excel - i.e. the original note is not removed unless pasted somewhere else. Just a thought.

Issue #21, opened 2017-02-25
Zooming/focusing into bullets


neomodern:
I love deep nesting in an outliner, and this feature requires being able to hide all the other notes on a page and full-screen the notes I'm focusing on. 

Ideally this is implemented in a similar style to WorkFlowy.

a13ph:
In emacs/orgmode this is called narrowing, btw

On Feb 26, 2017 1:55 AM, "neomodern" <notifications@github.com> wrote:

I love deep nesting in an outliner, and this feature requires being able to
hide all the other notes on a page and full-screen the notes I'm focusing
on.

Ideally this is implemented in a similar style to WorkFlowy.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/orgzly/orgzly-android/issues/21>, or mute the thread

jhegedus42:
brilliant idea, i had the same thought after i realized that the screen is running out really fast on my android

orgzly is amazing already, that feature would make it unbeatable on phones

arvinboggs:
I'd to request this (narrowing) feature as well. This is the feature lacking in orgzly that prevents me from using it. Currently, I'm using MindNote.

ajparkme:
This feature request gets my vote for sure.

danielo515:
I want to add that this is also very handy when you want to rapidly add several entries under the same section

AB1908:
I'd also love this. I have some deep nesting and it's easy to get lost.

PR #319, opened 2018-04-26
Org refile

karlicoss:
I actually imagined it would be complicated, but apparently existing cut and paste work perfectly for that. Hopefully inserting under root node is not too hacky? 

Also I'm not a great designer, so feel free to propose position in menu and icon (if you fancy it, otherwise I'd find some appropriate icon under free license).

I also think it would be useful to have refile action in swipe menu? Left/right? Let me know and I'll add it too.

The bit which I have not implemented properly yet is in `openNoteRefileDialog` -- I just quickly hacked it and displaying notebook names in alert dialog. However, perhaps we want to reuse `BooksFragment` somehow? I didn't do it in this PR because first, `BooksFragment` is not `DialogFragment`, so might require a wrapper fragment to show it as dialog, and second, it displays some additional information (modification date, sync status), which we probably don't want in refile menu which ought to be quick and compact. Anyway, let me know how do you vision this so I'd implement it properly.

I'd imagine this PR would resolve https://github.com/orgzly/orgzly-android/issues/95

In the "Test for moving note down commit" I was just familirizing myself to orgzly source code, hopefully that would be useful too.

P.S. Great job, could easily be best open source android app I've seen so far; everything is well structured, plenty of tests -- pleasure to contribute!


nevenz:
> Hopefully inserting under root node is not too hacky?

I think it's the way to go.

> Also I'm not a great designer, so feel free to propose position in menu and icon

Most icons are from [google/material-design-icons](https://github.com/google/material-design-icons). For example, I'd use [this one](https://github.com/google/material-design-icons/blob/master/action/drawable-xxxhdpi/ic_open_in_browser_black_24dp.png) for refiling, though there might be even more appropriate one.

As for position, yeah, I think it should definitely go to quick-menu too. I'd say last icon in left swipe menu (as those are the ones that change the notebook structure).

> The bit which I have not implemented properly yet is in openNoteRefileDialog

Originally, I was planning to use `BrowserFragment` and implement note browsing (like `FileBrowserFragment` implements file browsing). But I think dialog would be better. I don't think `BooksFragment` could be used, as this will need to support refiling under a specific note too, not just a notebook. Still, it does scream "code reuse", as you'd be browsing notes for refile the same way you do now.

> I'd imagine this PR would resolve #95

Sounds like it, though personally I never refile to the end of notebook. There could be a different issue for refiling under specific note, so that we don't try and do everything at once.

One bug: refiling multiple notes. Try selecting two on a different level, from different branches. Cutting would need to normalize those first and sort out all the values in `NotePosition`. This is a big PITA and the reason why cutting&pasting is not yet done for multiple notes. So short-term you can just disable the action if multiple notes are selected.

Thanks!

karlicoss:
Good catch, completely forgot about refiling multiple notes, good catch. I guess because never do that in emacs.

I think all done!

Also, was thinking -- potentially it might be useful to refile notes from search. However, in search you don't easily have `bookId`, so I didn't bother with that at this point. However, after a bit of thinking, I realized I don't understand why would cut operation need `bookId`. It seems that it's needed in `whereDescendantsAndNotes` and `whereAncestors` queries, but i think for these queries it would be enough to have `a.bookId = b.bookId` condition, since book is determined uniquely by the note ids anyway. That way cut operation would be independent of book id, so you could cut and refile from search/agenda views.

nevenz:
> Also, was thinking -- potentially it might be useful to refile notes from search.

My main concern for doing any structure modification from search results is that you're kinda doing it blind. Search results are a "virtual view". Note could have children for example and most actions would have to be applied to them as well, which might not be what you want or/and this would have to be made obvious.

> However, after a bit of thinking, I realized I don't understand why would cut operation need bookId.

There's also updating of book's modification time. You could certainly get book id with one more query, or use a subquery when needed, though it would make things more complicated.

BTW, we probably want to make notebook in note view clickable with this action, but that could be done separately from this PR.

karlicoss:
> you're kinda doing it blind

Right, can see your point. Could potentially have some logic for prompting user in case it hase notes/children, but it's more comlicated.

> There's also updating of book's modification time

ah ok can see it now

> we probably want to make notebook in note view clickable with this action

Good idea. But wouldn't that potentially suffer from the 'doing it blind' argument? Since you can't see note's children in note view.
Anyway I gave it a quick shot; it's fairly easy to make notebook clickable, it already is in `ShareActivity` and the code can easily be reused. However it's gonna be harder to make it consistent with note state change logic (pressing back cancels all modifications). I adapted it via including notebook id in the note state; but right now note updating from `NoteFragment` works on single note only, so child notes would get orphaned unless you use do something more complicated.
So maybe giving it more thought and doing in a separate PR would indeed be a good idea.

nevenz:
> Good idea. But wouldn't that potentially suffer from the 'doing it blind' argument? Since you can't see note's children in note view.

Until #21 is implemented, yes.

> However it's gonna be harder to make it consistent with note state change logic (pressing back cancels all modifications). I adapted it via including notebook id in the note state; but right now note updating from NoteFragment works on single note only, so child notes would get orphaned unless you use do something more complicated.

I imagine it would be similar to the call to `onNoteCreateRequest`.

PR #338, opened 2018-05-24
Correct link on how to disable animations.

bjonnh:
The link was shortened and was redirecting to a non-existent page.


nevenz:
Thanks!

Issue #58, opened 2017-04-01
Support checkboxes


wpolitarczyk:
Org-mode supports creating checklists which is very nice and useful feature. It would be really nice if Orgzly could also handle checkboxes.

iknownothingaboutcomputers:
And automatically move the checked items to the bottom of the list.

PR #342, opened 2018-06-03
Add possibility to check/uncheck checkboxes

halvorlu:
Issue #58 

This is a suggested implementation of checkboxes, which probably needs some polishing before merging. I'm creating the pull request to get some feedback/advice for (possibly) improving the implementation.

The checkboxes work as follows:
- When you edit the content of a note, "c " at the beginning of a line is automatically replaced by "[ ] ". This is to make it easier to write checkboxes (easier than writing brackets).
- When you view a note, the checkboxes are shown in monospace font, green if "[X]" and red if "[ ]".
- Checkboxes can be checked/unchecked by clicking them.
- It seems that some other formatting (bold etc.) is lost as it is now. I need to fix this.

Possible improvements:
- Using the same exact color for checkboxes as for TODO/DONE
- Using a proper Android Checkbox when viewing the checkbox?
- Using a different trigger for replacement than "c "?
- An option to turn on/off the trigger text, or even edit it?
- Add a checkbox example to the "Getting Started with Orgzly" notebook

### DIFF
--- a/app/src/main/java/com/orgzly/android/ui/HeadsListViewAdapter.java
+++ b/app/src/main/java/com/orgzly/android/ui/HeadsListViewAdapter.java
@@ -225,6 +225,17 @@ public class HeadsListViewAdapter extends SimpleCursorAdapter {
                     // TODO: How to update the note without causing a recursion?
                     //NotesClient.update(context, note);
                     holder.content.addTextChangedListener(this);
+
+                    // Consider doing this?
+                    /*
+                    if (BuildConfig.LOG_DEBUG) LogUtils.d(TAG, "Text modified in list", s);
+                    ActionService.Companion.enqueueWork(
+                            context,
+                            new Intent(context, ActionService.class)
+                                    .setAction(AppIntent.ACTION_UPDATE_NOTE)
+                                    .putExtra(AppIntent.EXTRA_NOTE_CONTENT, s.toString()));
+                                        */
+
                 }
             });
 
--- a/app/src/main/java/com/orgzly/android/ui/HeadsListViewAdapter.java
+++ b/app/src/main/java/com/orgzly/android/ui/HeadsListViewAdapter.java
@@ -8,6 +8,8 @@ import android.graphics.Typeface;
 import android.graphics.drawable.Drawable;
 import android.os.AsyncTask;
 import android.support.v4.widget.SimpleCursorAdapter;
+import android.text.Editable;
+import android.text.TextWatcher;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.LinearLayout;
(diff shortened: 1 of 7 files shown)


nevenz:
I've done something similar recently, I should have committed it earlier. I've pushed it now to https://github.com/orgzly/orgzly-android/commit/534759c5d796f12704aab3a3a5ea9a0b64bcdbc8

So check if there's something you can use. Perhaps a pattern and updating note though `ActionService`.

> When you edit the content of a note, "c " at the beginning of a line is automatically replaced by "[ ] ". This is to make it easier to write checkboxes (easier than writing brackets).
> Using a different trigger for replacement than "c "?

I don't know if it's worth supporting this for checkbox, I think it would be better to add a toolbar on the top and have buttons for checkboxes, bold, italic etc. Though it's an interesting idea in general.

What I think it would be useful is pressing enter at the end of the like with checkbox and creating a new one below.

halvorlu:
Thanks for the input. I've done some more changes now:
- Removed the "c " trigger text
- Checkboxes are now automatically added if the user presses enter on the end of a checkbox line. This also works with indented checkboxes.
- I've added a new button next to "View/Edit content" to insert checkbox, just as a suggestion. Not sure if it is the best solution. The automatic checkbox on Enter may be enough?
- I still haven't figured out how to update the note if checkboxes are clicked when viewing the whole notebook.
- I think it would be really useful to have the checkboxes colored as TODO/DONE. Haven't found out how to access the correct colors, could you help here?

dstrelnikov:
If somebody want to hear me, I'm very vouting on the
> Using a proper Android Checkbox when viewing the checkbox?

Because Orgzly, after all, is a **mobile** realisation of org mode. I have already true Emacs on my smartphone inside of Termux. So, I think, Orgzly should be more touch-like. 

How I see this? When you have a list in a note, this list is displayed separately, like the metadata of the note. Or, for example, by tapping on the list, you get a pop-up window like window with todo states.

memeplex:
Isn't it a bit inconsistent that only checklists support this "electric mode" that inserts a new item after pressing enter? I understand that they are more difficult to type than other kind of lists, and maybe that they are most useful (but this depends on the context, reference notes vs task lists), but nevertheless this change may lead to the unfulfilled expectation of auto-itemized lists in general.  

Analogously asterisks in lists are currently not formatted as bullets, hence formatting checkboxes could feel ad hoc.

memeplex:
And similarly for description lists, although this could be too much.

nevenz:
> Isn't it a bit inconsistent that only checklists support this "electric mode" that inserts a new item after pressing enter?

It can be removed after adding toolbar. Especially if it's located right above keyboard. Pressing button for checklist item instead of enter wouldn't be slower.

halvorlu:
We could consider implementing automatic insertion of a new item with dash/number/checkbox when pressing enter in a new list, so that it applies to all lists, not only those with checkboxes?

nevenz:
> Perhaps TextViewWithMarkup could have its own markupTextChangedListener (or similar), that would get called after toggling checkboxes, folding drawers etc.

I've done this in 6e0fe94c.

I've also removed coloring and that button in `NoteFragment` for now.

Thanks again!

PR #363, opened 2018-07-11
Add "quick reply" to notification as a way to add new notes

czan:
When entering notes I want to be able to quickly write something and store it away to deal with later. At the moment this involves me opening a screen with lots of fields, filling in the one field I care about (the title), then saving the note. I would much rather use the "quick reply" feature on the notification to be able to enter a note, without interrupting what I am currently doing. This is similar to the way I use `org-capture`.

This PR contains code to implement this. I'm not _completely_ happy with it, but I like to start a conversation with a proof of concept. I'm not familiar with how Orgzly works, so it's likely that I have not done this the right way.

### DIFF
--- a/app/src/main/java/com/orgzly/android/NewNoteBroadcastReceiver.java
+++ b/app/src/main/java/com/orgzly/android/NewNoteBroadcastReceiver.java
@@ -1,0 +1,47 @@
+package com.orgzly.android;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+import android.support.v4.app.RemoteInput;
+
+import com.orgzly.android.prefs.AppPreferences;
+import com.orgzly.android.provider.clients.NotesClient;
+import com.orgzly.org.OrgHead;
+
+import java.io.IOException;
+
+public class NewNoteBroadcastReceiver extends BroadcastReceiver {
+    public static final String NOTE_TITLE = "NOTE_TITLE";
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        String title = getNoteTitle(intent);
+        if (title != null) {
+            try {
+                Book targetBook = BookUtils.getTargetBook(context);
+                NotesClient.create(context, makeNote(AppPreferences.newNoteState(context), title, targetBook));
+                Notifications.createNewNoteNotification(context);
+            } catch (IOException ex) {
+                ex.printStackTrace();
+            }
+        }
+    }
+
+    private Note makeNote(String state, String title, Book targetBook) {
+        OrgHead head = new OrgHead();
+        head.setTitle(title);
+        head.setState(state);
+        Note note = new Note();
+        note.setCreatedAt(System.currentTimeMillis());
+        note.getPosition().setBookId(targetBook.getId());
+        note.setHead(head);
+        return note;
+    }
+
+    private String getNoteTitle(Intent intent) {
+        Bundle remoteInput = RemoteInput.getResultsFromIntent(intent);
+        return remoteInput == null ? null : remoteInput.getString(NOTE_TITLE);
+    }
+}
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/orgzly/android/Notifications.java
```diff
@@ -45,6 +49,22 @@ public static void createNewNoteNotification(Context context) {
                 .setContentIntent(resultPendingIntent)
                 .setPriority(NotificationCompat.PRIORITY_MIN); // Don't show icon on status bar
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT_WATCH) {
```
nevenz: It seems this needs to be `Build.VERSION_CODES.N` (https://developer.android.com/training/notify-user/build-notification#reply-action). Even though `RemoteInput` has been added earlier. Also tested on API 23 emulator, where nothing happens when you click the button.

app/src/main/java/com/orgzly/android/NewNoteBroadcastReceiver.java
```diff
@@ -0,0 +1,67 @@
...
+        String title = getNoteTitle(intent);
+        Book targetBook = getTargetBook(context);
+        if (title != null && targetBook != null) {
+            NotesClient.create(context, makeNote(AppPreferences.newNoteState(context), title, targetBook));
+            Notifications.createNewNoteNotification(context);
+        }
+    }
+
+    private Note makeNote(String state, String title, Book targetBook) {
+        Note note = Note.newRootNote(targetBook.getId());
```
nevenz: "Root note" is a dummy note, it's not visible and has level 0. This works because those values are overwritten when note is added to the book. But just to avoid confusion, we probably want something like:

```
Note note = new Note();
note.getPosition().setBookId(targetBook.getId());
```

app/src/main/java/com/orgzly/android/NewNoteBroadcastReceiver.java
```diff
@@ -0,0 +1,67 @@
...
+        note.setCreatedAt(System.currentTimeMillis());
+        return note;
+    }
+
+    private String getNoteTitle(Intent intent) {
+        Bundle remoteInput = RemoteInput.getResultsFromIntent(intent);
+        return remoteInput == null ? null : remoteInput.getString(NOTE_TITLE);
+    }
+
+    private Book getTargetBook(Context context) {
```
nevenz: This is the same thing from `ShareActivity`, right? Maybe it can be moved to `BookUtils` for example, and reused?

nevenz:
This is cool, thanks! I've added few comments, but it looks good to me.

Perhaps button could be renamed to "Capture", to make it more distinct from the the main action which opens the app?

app/src/main/java/com/orgzly/android/Notifications.java
czan: Ah, good one. I should have tested that myself, but I only tested 19 and 27, I think. I'll fix that up, thanks!

app/src/main/java/com/orgzly/android/NewNoteBroadcastReceiver.java
czan: Yep, it is. Definitely a better idea for it to be in `BookUtils`. I'll move it.

czan:
My main argument against "Capture" is that it might not be clear exactly what that button does to people who aren't already familiar with `org-capture`. Maybe it could be something like "Quick note"?

nevenz:
> My main argument against "Capture" is that it might not be clear exactly what that button does to people who aren't already familiar with org-capture. Maybe it could be something like "Quick note"?

Yeah, that's true. Sure, "Quick note" sounds fine.

czan:
I've just pushed new changes, to address your comments. I unthinkingly just modified the history, but hopefully that won't cause any issues. Given it's still just a PR I assume it will be fine.

nevenz:
Thanks again!

Issue #360, opened 2018-07-04
file links


steckerhalter:
org supports file links like this: `[[../stuff/here.mp3]]` for example
could this be implemented as well? so it would just open the file?

I'm submitting a **feature request**.

Ypot:
But wouldn't it be a little bit weird, since  the paths from the app would be different from the ones in orgmode?

Could it be feasible to stablish LAN Network paths? So there could be a coherence between linked files in orgmode and orgzly, at least at home.

steckerhalter:
they wouldn't be different on my system. it should just do what orgmode does IMO. anything else would be an additional feature

speakerToSystems:
I'd like this too.
I want to use orgzly for project notes, with document links.
While it's be great to have some kind of portable link syntax, or a mode specific link syntax (substitute this file link on Android for the regular orgmode link), I'd be happy enough just having 2 links to each file, 1 for orgmode and 1 for Android.

steckerhalter:
hey, thanks a lot for implementing this...

speakerToSystems:
Thanks for implementing this.
I'm not quite there yet, though.
I'm trying to link to a file on my SD card.
Here's my link:
[[file:/storage/extSdCard/Android/data/com.ghisler.android.TotalCommander/files/t.pdf]]
I get an error message starting with
Failed to find configured root that contains /storage/emulated/0/Android/data/...

nevenz:
> [[file:/storage/extSdCard/Android/data/com.ghisler.android.TotalCommander/files/t.pdf]]
> I get an error message starting with
> Failed to find configured root that contains /storage/emulated/0/Android/data/...

See https://github.com/orgzly/orgzly-android/pull/389#issuecomment-448546202

lingnand:
This doesn't seem to work for url encoded file link e.g., `folder/File%20With%20Space.pdf`. If I have replace `%20` with literal space that seems to work though - is this the expected behavior?

lingnand:
This doesn't seem to work for url encoded file link e.g., `folder/File%20With%20Space.pdf`. If I have replace `%20` with literal space that seems to work though - is this the expected behavior?

lingnand:
This doesn't seem to work for url encoded file link e.g., `folder/File%20With%20Space.pdf`. If I have replace `%20` with literal space that seems to work though - is this the expected behavior?

nevenz:
[...]

PR #382, opened 2018-09-05
Handle "file:" links for opening files on android storage

aancel:
Hello,

This is an attempt at implementing file links within orgzly (linked to #360).
Fow now, it will only work with the `file:path` link form, but tuning the regex might enable other forms.

This MR uses the file provider of Android, which should apparently be the way to go since Android 7.0. It tries to find the file links in the content, creates a `content://` link for a file in the internal storage (Right now the base path is /sdcard) and finally tries to open the file with an `ACTION_VIEW` intent.

This seems to be working quite fine, as I tested it and was able to open png, jpg and pdf files (on the Pixel 2 XL emulator with Android 8.1).

Things that won't be working as of now:
- If you have a space in the path to your file, the regex matcher will stop at it, preventing you to get the full path

This is my first time working with kotlin, so there might be mistakes.

Also thanks to @nevenz for this great app and to @halvorlu for his MR #342, that I followed and it helped me in trying to implement this.

### DIFF
--- a/app/src/main/java/com/orgzly/android/ui/CommonActivityLifecycleCallbacks.java
+++ b/app/src/main/java/com/orgzly/android/ui/CommonActivityLifecycleCallbacks.java
@@ -22,12 +22,12 @@ public class CommonActivityLifecycleCallbacks  implements Application.ActivityLi
 
     @Override
     public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
-        registerActivity(activity);
+        registerActivity(null);
     }
 
     @Override
     public void onActivityStarted(Activity activity) {
-        registerActivity(activity);
+        registerActivity(null);
     }
 
     @Override
--- a/app/src/main/java/com/orgzly/android/ui/CommonActivityLifecycleCallbacks.java
+++ b/app/src/main/java/com/orgzly/android/ui/CommonActivityLifecycleCallbacks.java
@@ -1,0 +1,57 @@
+package com.orgzly.android.ui;
+
+import android.app.Activity;
+import android.app.Application;
+import android.os.Bundle;
+
+import com.orgzly.android.App;
+
+// Monitor the lifecycle of CommonActivity
+// This allows to access the activity in files like OrgFormatter
+// and enables permission and snackbar access
+public class CommonActivityLifecycleCallbacks  implements Application.ActivityLifecycleCallbacks {
+
+    public void registerActivity(Activity activity)
+    {
+        if (activity instanceof CommonActivity) {
+            App.setCurrentActivity((CommonActivity) activity);
+        } else {
+            App.setCurrentActivity(null);
+        }
+    }
+
+    @Override
+    public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
+        registerActivity(activity);
+    }
+
+    @Override
+    public void onActivityStarted(Activity activity) {
+        registerActivity(activity);
+    }
+
+    @Override
+    public void onActivityResumed(Activity activity) {
+        registerActivity(activity);
+    }
+
+    @Override
+    public void onActivityPaused(Activity activity) {
+        registerActivity(null);
+    }
+
+    @Override
+    public void onActivitySaveInstanceState(Activity activity, Bundle outState) {
+        registerActivity(null);
+    }
+
+    @Override
+    public void onActivityStopped(Activity activity) {
+        registerActivity(null);
+    }
+
+    @Override
+    public void onActivityDestroyed(Activity activity) {
+        registerActivity(null);
+    }
+}
--- a/app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
+++ b/app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
@@ -4,7 +4,6 @@ import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
 import android.graphics.Typeface
-import android.net.Uri
 import android.os.Environment
 import android.support.v4.content.ContextCompat.startActivity
 import android.text.SpannableString
@@ -21,6 +20,7 @@ import java.util.regex.Matcher
 import java.util.regex.Pattern
 import android.support.v4.content.FileProvider
 import com.orgzly.BuildConfig
+import com.orgzly.R
 import java.io.File
 
 
@@ -370,28 +370,37 @@ object OrgFormatter {
                     override fun onClick(widget: View?) {
                         val context = App.getAppContext()
 
-                        // Get the file
-                        val file = File(Environment.getExternalStorageDirectory().getPath() + "/" + s.get(1))
+                        // Check that we have the permission to read external files
+                        val isGranted = AppPermissions.isGranted(context, AppPermissions.Usage.EXTERNAL_FILES_ACCESS)
 
-                        // Check file existence, before trying to process it
-                        if(file.exists()) {
-                            val contentUri = FileProvider.getUriForFile(context, BuildConfig.APPLICATION_ID + ".fileprovider", file)
+                        if (isGranted) {
+                            // Get the file
+                            val file = File(Environment.getExternalStorageDirectory(), s[1])
 
-                            val intent = Intent(Intent.ACTION_VIEW, Uri.parse(contentUri.toString()))
-                            // Added for support on API 16
-                            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
-                            intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+                            // Check file existence, before trying to process it
+                            if(file.exists()) {
+                                val contentUri = FileProvider.getUriForFile(context, BuildConfig.APPLICATION_ID + ".fileprovider", file)
 
-                            // Try to start an activity for opening the file
-                            try {
-                                startActivity(context, intent, null)
-                            } catch (e: ActivityNotFoundException) {
-                                Toast.makeText(context, "No application found to open this file format", Toast.LENGTH_SHORT).show()
+                                val intent = Intent(Intent.ACTION_VIEW, contentUri)
+                                // Added for support on API 16
+                                intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                                intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+
+                                // Try to start an activity for opening the file
+                                try {
+                                    startActivity(context, intent, null)
+                                } catch (e: ActivityNotFoundException) {
+                                    Toast.makeText(context, R.string.external_file_no_app_found, Toast.LENGTH_SHORT).show()
+                                }
+                            }
+                            else
+                            {
+                                Toast.makeText(context, String.format(context.getString(R.string.external_file_not_found), file.absolutePath), Toast.LENGTH_SHORT).show()
                             }
                         }
                         else
                         {
-                            Toast.makeText(context, "File " + file.absolutePath + " does not exist", Toast.LENGTH_SHORT).show()
+                            Toast.makeText(context, R.string.permissions_rationale_for_external_files_access, Toast.LENGTH_SHORT).show()
                         }
                     }
                 }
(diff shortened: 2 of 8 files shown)


nevenz:
I can't seem to get any link to work. Can you give an example of a link and where the file should go?

I was playing with:

```
file:asd

file:song.mp3
file:logo.jpg
file:document.txt

file:Download/song.mp3
file:Download/logo.jpg
file:Download/document.txt

file:non-existing/song.mp3
file:non-existing/logo.jpg
file:non-existing/document.txt
```

I get a crash for the first one:

```
09-06 10:25:05.212 14039 14039 E MessageQueue-JNI: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=content://com.orgzly.fileprovider/sdcard/asd flg=0x1 }
09-06 10:25:05.212 14039 14039 E MessageQueue-JNI:      at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:1937)
09-06 10:25:05.212 14039 14039 E MessageQueue-JNI:      at android.app.Instrumentation.execStartActivity(Instrumentation.java:1616)
09-06 10:25:05.212 14039 14039 E MessageQueue-JNI:      at android.app.ContextImpl.startActivity(ContextImpl.java:878)
09-06 10:25:05.212 14039 14039 E MessageQueue-JNI:      at android.content.ContextWrapper.startActivity(ContextWrapper.java:394)
09-06 10:25:05.212 14039 14039 E MessageQueue-JNI:      at android.support.v4.content.ContextCompat.startActivity(ContextCompat.java:143)
09-06 10:25:05.212 14039 14039 E MessageQueue-JNI:      at com.orgzly.android.util.OrgFormatter$parseFilelinks$clickableSpan$1.onClick(OrgFormatter.kt:369)
```

I could install some text editor that tries to open just about anything, avoiding the crash, but we can't rely on that of course.

Clicking links that point to existing files (in `/sdcard/<file>`) doesn't work for me:

For `song.mp3` I get a toast `Couldn't play the track you requested.` Opening that file from Files app (pre-installed app on emulator) works.

For `logo.jpg` viewer doesn't display anything. It works from Files app.

Opening `document.txt` with Chrome opens:
```
content://com.orgzly.fileprovider/sdcard/Download/document.txt
```
but the browser displays `Your file was not found`. Opening the same file from Files app opens:
```
content://com.android.providers.downloads.documents/document/raw%3A%2Fstorage%2Femulated%2F0%2FDownload%2Fdocument.txt
```
and the file content is displayed.

So I think this displays the issue -- I don't think URL is constructed properly. I don't think it should be created manually. I believe `FileProvider.html#getUriForFile` might be a way to go?

Also, on emulator with API 16 I get:

```
E/MessageQueue-JNI( 2101): android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag. Is this really what you want?
E/MessageQueue-JNI( 2101):      at android.app.ContextImpl.startActivity(ContextImpl.java:886)
E/MessageQueue-JNI( 2101):      at android.content.ContextWrapper.startActivity(ContextWrapper.java:281)
E/MessageQueue-JNI( 2101):      at android.support.v4.content.ContextCompat.startActivity(ContextCompat.java:143)
E/MessageQueue-JNI( 2101):      at com.orgzly.android.util.OrgFormatter$parseFilelinks$clickableSpan$1.onClick(OrgFormatter.kt:369
```

for any link. I don't know if this is just not supported (you mention Android 7.0) and needs to be handled, or it's done differently there.

Thanks!

aancel:
I tried the files in a similar setup as you did:
```
file:Download/song.mp3
file:Download/sample.pdf
file:Download/test.png
```

All files where successfully opened, either via Google Music, Google Drive or Google Photos apps (I am only using the emulator here with a Pixel XL on Android 8.1)

Indeed for the crash, we can't rely on users installing specific apps for this specific purpose. 
Maybe there is a way to check if an app is available to open the file that is being opened, I'll have a look !

I also agree on the fact that the file name is not really being properly built for now. I'll have a look at  `FileProvider.html#getUriForFile` to see if it helps.
I followed [this](https://stackoverflow.com/questions/38200282/android-os-fileuriexposedexception-file-storage-emulated-0-test-txt-exposed) discussion regarding the use of FileProvider. There are some hacky ways to avoid the restriction to access the files on the device in place since android N apparently, but I'd prefer using the FileProvider (unless you don't mind the other ways ?)

I'll try to use the app on other versions of android to see if I can reproduce the errors that you encountered.

Thanks for the feedback !

aancel:
I have updated the code with some additional checks, notably checking for file existence and catching the exception when no application can be found for opening the external file.
Support for API 16 should also be ok, with the extra Intent flag I added.

I have tested this (only via the emulator) with: API 16, API 23 and API 27

One thing that I thought while working on this is that we could use the `updateDrawState` overridable method of the `ClickableSpan` to indicate if a file exists, like setting the link to red if the file cannot be found and the normal color if it is found.
What do you think about this ?

Hopefully this helps solving the issues you encountered, or at least prevent them !

nevenz:
> I have updated the code with some additional checks, notably checking for file existence and catching the exception when no application can be found for opening the external file.

Cool, I think just catching exception is fine.

> Support for API 16 should also be ok, with the extra Intent flag I added.

Works for me now.

> I have tested this (only via the emulator) with: API 16, API 23 and API 27

I still see the same. `file.exists()` does pass, but I don't get any of the files opened. URL is the same BTW, `FileProvider.getUriForFile` didn't change anything.

I tried emulator with API 27 (Pixel 2) and API 23 (Nexus 5X). Both after wiping the data just in case.

> One thing that I thought while working on this is that we could use the updateDrawState overridable method of the ClickableSpan to indicate if a file exists, like setting the link to red if the file cannot be found and the normal color if it is found.

Sounds cool, but I'm afraid it might be too slow, especially in book view.

aancel:
I think that I might have pinpointed the issue, following a data wipe as suggested.
Wiping the data indeed ended for me as the file not opening properly while still going through the `file.exists()` test.

I think that the issue is related to the Storage permission being reset for the Orgzly app.
When I do not allow Orgzly to access the storage, and try to open an image, it tries to open it by launching an activity, but the image is not displayed. Allowing Orgzly access to storage fixes the opening issue.
Can you confirm this on your side ?

If it is indeed the issue, what do you think about asking for the storage permission, if it is not already given, when the user clicks on a file link ?
I see that you already have an AppPermission.kt file for this, should I use this code ?

Regarding the link color change when the file is badly linked, I was also afraid about performance.
There is already a toast indicating that the file does not exist, so it might be enough.
(By the way, are you ok with the Toast ? Or would you have preferred another way ?)

nevenz:
> Allowing Orgzly access to storage fixes the opening issue.

Ah, yes it works when I grant Storage permission to the app. But I thought the whole point of using uri-based permissions (`grantUriPermissions` and `FLAG_GRANT_READ_URI_PERMISSION`) is to avoid that.

> If it is indeed the issue, what do you think about asking for the storage permission, if it is not already given, when the user clicks on a file link ?
> I see that you already have an AppPermission.kt file for this, should I use this code ?

See how `CommonActivity#runWithPermission` is used. Though not having activity available might be an issue.

> By the way, are you ok with the Toast ? Or would you have preferred another way ?

There is only only place where Toast is used now in Orgzly (when activity is immediately terminated afterwards). Snackbars are used everywhere else. But it might not be possible to use them here because activity is not available (same issue as above).

aancel:
> Ah, yes it works when I grant Storage permission to the app. But I thought the whole point of using uri-based permissions (grantUriPermissions and FLAG_GRANT_READ_URI_PERMISSION) is to avoid that.

Regarding this, I think that the uri-based permission is required for the launched activity, but Orgzly still needs to have access to the storage apparently.

> See how CommonActivity#runWithPermission is used. Though not having activity available might be an issue.

I've updated the code for the permissions. As you said, I didn't find a reliable/clean way to access to the Activity. For now, the app can only check if the permission is granted (and inform the user if not), but not request it, which is a bit annoying.
I've also moved the strings of the toasts into the `string.xml` file.

The code has also been updated according to your reviews

nevenz:
> Regarding this, I think that the uri-based permission is required for the launched activity, but Orgzly still needs to have access to the storage apparently.

OK, that makes sense I guess.

> As you said, I didn't find a reliable/clean way to access to the Activity.

It seems one way to do it is to use `Application#registerActivityLifecycleCallbacks` in `App#onCreate`, keeping track of current activity:

```
registerActivityLifecycleCallbacks(new ActivityLifecycleCallbacks() {
            // ...

            @Override
            public void onActivityResumed(Activity activity) {
                if (activity instanceof CommonActivity) {
                    currentActivity = (CommonActivity) activity;
                } else {
                    currentActivity = null;
                }
            }

            @Override
            public void onActivityPaused(Activity activity) {
                currentActivity = null;
            }

            // ...
        });
```

Then it would be possible to use `App.currentActivity?.runWithPermission` and `App.currentActivity?.showSnackbar`.

aancel:
I've updated the code following your advice with the `registerActivityLifecycleCallbacks` method.
Hopefully eveything is placed as you would have wanted.

Now, Orgzly will use snackbars to display the messages (previoulsy displayed as Toasts) and it will ask for storage permission, when clicking a link and if it has not already been granted.
Also, I had to rework the logic for dismissing the snackbar whith the UP touch event. Using the previous implementation, the snackbars appeared and were immediately dismissed when clicking on a link. I've tested the new code with other snackbars (like the sync repository), and it seem to behave as it was before !

app/src/main/java/com/orgzly/android/ui/CommonActivityLifecycleCallbacks.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public void onActivityStarted(Activity activity) {
+        registerActivity(activity);
+    }
+
+    @Override
+    public void onActivityResumed(Activity activity) {
+        registerActivity(activity);
```
nevenz: I think this is the only place where activity should be set, just to be safe.

app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
```diff
@@ -332,6 +347,68 @@ object OrgFormatter {
...
+                        val context = App.getAppContext()
+
+                        // Get the current activity is available
+                        val currentActivity = App.getCurrentActivity()
+                        if(currentActivity != null) {
+                            // Check that we have the permission to read external files
+                            // or ask for it
+                            val isGranted = AppPermissions.isGrantedOrRequest(currentActivity, AppPermissions.Usage.EXTERNAL_FILES_ACCESS)
+
+                            if (isGranted) {
```
nevenz: After granting permission, user needs to click the link again. Check `CommonActivity#runWithPermission` usages. It'll run `Runnable` again after permission is granted.

app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
```diff
@@ -332,6 +347,68 @@ object OrgFormatter {
...
+                                    intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+
+                                    // Try to start an activity for opening the file
+                                    try {
+                                        startActivity(context, intent, null)
+                                    } catch (e: ActivityNotFoundException) {
+                                        currentActivity.showSnackbar(R.string.external_file_no_app_found)
+                                    }
+                                } else {
+                                    currentActivity.showSnackbar(String.format(context.getString(R.string.external_file_not_found), file.absolutePath))
```
nevenz: `context.getString` accepts arguments, so message can be:
```
context.getString(R.string.external_file_not_found, file.absolutePath)
```

aancel:
I've updated the code according to your reviews.
Please tell if anything else can be improved !

nevenz:
In future, we'll probably want to allow user to configure a base directory. Using a repository (if `Directory`) by default can be an option too, together with syncing other non-org files optionally.

Thanks again!

nevenz:
> Also, I had to rework the logic for dismissing the snackbar whith the UP touch event. Using the previous implementation, the snackbars appeared and were immediately dismissed when clicking on a link. I've tested the new code with other snackbars (like the sync repository), and it seem to behave as it was before !

I though I tested this, but clicking on action button doesn't seem to work ("Repositories" after Sync with no repos for example).

Issue #147, opened 2017-08-13
Display images


crocefisso:
In my Orgzly folder, I created a Pic folder to store pictures that I'd like to display for some todos. In order to display picture P on some note, I input [[./Pic/P.jpg]] and [[./Pic/P.jpg][Picture P] in my note but neither make displaying available. Adding this feature would be useful.

ookimi:
This would indeed be great. I'm using org-mode with the org-drill for study notes, and pictures are often needed. In combination with the timestamps and tags, pictures in orgzly would make reviewing a breeze.

mrDSide:
Maybe would be nice to have feature like in desktop version for categories?
https://julien.danjou.info/icon-category-support-in-org-mode/

vzaliva:
One more vote for pictures. I often take pictures of whiteboards during my meeting and refer to them in the notes. Currently, I keep them in the same Dropbox folder as my org-mode meeting minutes and emacs shows them just fine. It would be nice for orgzly to show them as well.

Ypot:
That would be superb. We could take notes fast using even photos.

nevenz:
Was done in #389.

Issue #348, opened 2018-06-16
Broken UI actions after settings


fairplay:
I'm submitting a **bug report** for version 1.6.7
After visiting settings and changing something (still not sure if it is a necessary condition, but I could reproduce the behaviour clicking "Force UTF-8" setting), then returning to the notebook via back key I've encountered the next problems:
- Back button is not working anymore
- Drawer can be displayed, but "Notebooks", "Settings" e.t.c. actions are not working anymore

nevenz:
I've seen it once too few versions ago, but haven't been able to reproduce it since. I suspect it's due to activity restart after leaving Settings.

If someone can reliably reproduce it, let me know.

aancel:
I am able to reproduce this issue with my setup in PR: #389 
We've managed to identify a potential cause in this comment: https://github.com/orgzly/orgzly-android/pull/389#issuecomment-428735893

@nevenz I did try the solution you proposed in https://github.com/orgzly/orgzly-android/pull/389#issuecomment-428737081, but this does not solve the issue. I still end up having 8 actions available instead of 14.
Could it be that a wrong instance of LocalBroadcastManager is updated instead ?

nevenz:
> Could it be that a wrong instance of LocalBroadcastManager is updated instead ?

No, it's a singleton.

Might be the issue of restarting the activity from `onResume` [here](https://github.com/orgzly/orgzly-android/blob/master/app/src/main/java/com/orgzly/android/ui/CommonActivity.kt#L216) maybe. You could try calling that from `Handler().postDelayed`, just to see what happens.

Also, I'd look for logs (or add some) and monitor `onResumeFragments|onReceive|onPause|onBookClicked` for example. When I return from settings, I see:
```
10-11 00:14:58.414 21865-21865/com.orgzly D/com.orgzly.android.ui.MainActivity: 2#main: onResumeFragments
10-11 00:14:58.493 21865-21865/com.orgzly D/com.orgzly.android.ui.MainActivity: 2#main: onPause
10-11 00:14:58.494 21865-21865/com.orgzly D/com.orgzly.android.ui.fragments.BookFragment: 2#main: onPause
10-11 00:14:58.575 21865-21865/com.orgzly D/com.orgzly.android.ui.MainActivity: 2#main: onResumeFragments
```

If you see `onPause` after `onResumeFragments`, it's a problem.

aancel:
I've first tried the logs solution.
[...]

PR #389, opened 2018-09-23
Allow "file:" and "[[...]]" links, display images in notes

aancel:
Hello,
Following the work on #382, I open a PR with new propositions.
I reworked the code a bit to better integrate it with the previous code, and notably to handle `[[...]]` for file links (still linked to #360) in addition to the previous `file:` scheme.
I also added an ImageSpan to display images in notes (rescaled to fit width, if larger than the window width), as a proposition to implement #147.

As before, this PR has been tested with API 16 (Nexus 5), API 23 (Pixel 2) & API 27 (Pixel 2 XL)

A small note sample, I used to test this:
```
[[Download/test.jpg]]

file:Download/test2.png

[[Download/test3.png][Test]]
```
Right now, named links are not replaced with an ImageSpan.
Furthemore regarding to the application performance, maybe adding a setting to toggle off image display could be a good idea. What do you think about this ?

### DIFF
--- a/app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
+++ b/app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
@@ -1,8 +1,24 @@
 package com.orgzly.android.ui
 
+import android.graphics.Color
+import android.graphics.drawable.ColorDrawable
+import android.graphics.drawable.Drawable
+import android.os.Environment
+import android.support.v4.content.FileProvider
 import android.text.Spannable
+import android.text.style.ImageSpan
+import android.view.View
+import com.orgzly.BuildConfig
+import com.orgzly.android.App
+import com.orgzly.android.prefs.AppPreferences
 import com.orgzly.android.ui.views.TextViewWithMarkup
 import com.orgzly.android.ui.views.style.FileLinkSpan
+import com.orgzly.android.util.AppPermissions
+import java.io.File
+import com.bumptech.glide.Glide
+import com.bumptech.glide.request.target.SimpleTarget
+import com.bumptech.glide.request.transition.Transition
+import com.bumptech.glide.request.RequestOptions
 
 
 object ImageLoader {
@@ -11,17 +27,116 @@ object ImageLoader {
         val context = textWithMarkup.context
 
         // Only if AppPreferences.displayImages(context) is true
-        // loadImages(textWithMarkup.text as Spannable)
-    }
-
-    private fun loadImages(text: Spannable) {
-        SpanUtils.forEachSpan(text, FileLinkSpan::class.java) { span ->
-            loadImage(span)
+        // Setup image visualization inside the note
+        if ( AppPreferences.displayInlineImages(context)
+                // Storage permission has been granted
+                && AppPermissions.isGranted(context, AppPermissions.Usage.EXTERNAL_FILES_ACCESS)) {
+            // Load the associated image for each FileLinkSpan
+            SpanUtils.forEachSpan(textWithMarkup.text as Spannable, FileLinkSpan::class.java) { span ->
+                loadImage(textWithMarkup, span)
+            }
         }
     }
 
-    private fun loadImage(span: FileLinkSpan) {
+    private fun loadImage(textWithMarkup: TextViewWithMarkup, span: FileLinkSpan) {
         val path = span.path
 
+        if (hasSupportedExtension(path)) {
+            val text = textWithMarkup.text as Spannable
+            // Get the current context
+            val context = App.getAppContext()
+
+            // Get the file
+            val file = File(Environment.getExternalStorageDirectory(), path)
+
+            if(file.exists()) {
+                // Get the Uri
+                val contentUri = FileProvider.getUriForFile(context, BuildConfig.APPLICATION_ID + ".fileprovider", file)
+
+                // Setup a placeholder
+                val drawable = ColorDrawable(Color.TRANSPARENT)
+                drawable.setBounds(0, 0, AppPreferences.setImageFixedWidth(context), AppPreferences.setImageFixedWidth(context))
+
+                Glide.with(context)
+                    .asDrawable()
+                    // Use a placeholder
+                    .apply(RequestOptions().placeholder(drawable))
+                    // And scaled thumbnails for faster display
+                    .thumbnail(Glide.with(context)
+                            .applyDefaultRequestOptions(RequestOptions().override(AppPreferences.setImageFixedWidth(context)))
+                            .load(contentUri))
+                    .load(contentUri)
+                    .into(object : SimpleTarget<Drawable>() {
+                        override fun onResourceReady(resource: Drawable, transition: Transition<in Drawable>?) {
+                            fitDrawable(textWithMarkup, resource)
+                            text.setSpan(ImageSpan(resource), text.getSpanStart(span), text.getSpanEnd(span), text.getSpanFlags(span))
+                        }
+                    })
+            }
+        }
+    }
+
+    fun hasSupportedExtension(path: String): Boolean {
+        var ret = false
+        var index: Int
+        var s = ""
+
+        // Find the last slash in the path
+        // to avoid case of a point in the path and a file without extension
+        index = path.lastIndexOf("/")
+
+        // If we found the last slash, extract the file name
+        if (index != -1) {
+            s = path.substring(index + 1)
+        }
+
+        // Extract the extension
+        index = s.lastIndexOf(".")
+
+        // If we found an extension, extract it and test it
+        if (index != -1) {
+            s = s.substring(index + 1).toLowerCase()
+
+            if (s == "jpg" || s == "jpeg" || s == "gif"
+                    || s == "png" || s == "bmp" || s == "webp") {
+                ret = true
+            }
+        }
+
+        return ret
+    }
+
+    fun fitDrawable(view: View, drawable: Drawable) {
+        // Get the display metrics to be able to rescale the image if needed
+        val metrics = view.context.resources.displayMetrics
+
+        // Gather drawable information
+        // Scale the height by the scaledDensity to get original image size
+        val drawableHeight = drawable.intrinsicHeight.toFloat() * metrics.scaledDensity
+        val drawableWidth = drawable.intrinsicWidth.toFloat() * metrics.scaledDensity
+
+        // Use either a fixed size or a scaled size according to user preferences
+        var fixedSize = -1
+        if (!AppPreferences.enableImageScaling(view.context)) {
+            fixedSize = AppPreferences.setImageFixedWidth(view.context)
+        }
+
+        if (fixedSize > 0) {
+            // Keep aspect ratio when using fixed size
+            val ratio = drawableHeight / drawableWidth
+            drawable.setBounds(0, 0, fixedSize, (fixedSize * ratio).toInt())
+        } else {
+            // Rescale the drawable if it is larger that the current view width
+            if (drawableWidth > view.width) {
+                //Compute image ratio
+                val ratio = drawableHeight / drawableWidth
+                // Ensure that the images have a minimum size
+                val width = Math.max(view.width, 256).toFloat()
+
+                drawable.setBounds(0, 0, width.toInt(), (width * ratio).toInt())
+            } else {
+                drawable.setBounds(0, 0, drawableWidth.toInt(), drawableHeight.toInt())
+            }
+        }
     }
 }
--- a/app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
+++ b/app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
@@ -1,8 +1,10 @@
 package com.orgzly.android.ui
 
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
 import android.graphics.Color
+import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.ColorDrawable
-import android.graphics.drawable.Drawable
 import android.os.Environment
 import android.support.v4.content.FileProvider
 import android.text.Spannable
@@ -53,23 +55,30 @@ object ImageLoader {
                 // Get the Uri
                 val contentUri = FileProvider.getUriForFile(context, BuildConfig.APPLICATION_ID + ".fileprovider", file)
 
+                // Get image sizes to reduce their memory footprint by rescaling
+                val options = BitmapFactory.Options()
+                options.inJustDecodeBounds = true
+                BitmapFactory.decodeFile(Environment.getExternalStorageDirectory().toString() + "/" + path, options)
+
+                val size = fitDrawable(textWithMarkup, options.outWidth, options.outHeight)
+
                 // Setup a placeholder
                 val drawable = ColorDrawable(Color.TRANSPARENT)
-                drawable.setBounds(0, 0, AppPreferences.setImageFixedWidth(context), AppPreferences.setImageFixedWidth(context))
+                drawable.setBounds(0, 0, size.first, size.second)
 
                 Glide.with(context)
-                    .asDrawable()
+                    .asBitmap()
                     // Use a placeholder
                     .apply(RequestOptions().placeholder(drawable))
-                    // And scaled thumbnails for faster display
-                    .thumbnail(Glide.with(context)
-                            .applyDefaultRequestOptions(RequestOptions().override(AppPreferences.setImageFixedWidth(context)))
-                            .load(contentUri))
+                    // Override the bitmap size, mainly used for big images
+                    // as it's useless to display more pixel that the pixel density allows
+                    .apply(RequestOptions().override(size.first, size.second))
                     .load(contentUri)
-                    .into(object : SimpleTarget<Drawable>() {
-                        override fun onResourceReady(resource: Drawable, transition: Transition<in Drawable>?) {
-                            fitDrawable(textWithMarkup, resource)
-                            text.setSpan(ImageSpan(resource), text.getSpanStart(span), text.getSpanEnd(span), text.getSpanFlags(span))
+                    .into(object : SimpleTarget<Bitmap>() {
+                        override fun onResourceReady(resource: Bitmap, transition: Transition<in Bitmap>?) {
+                            val bd = BitmapDrawable(App.getAppContext().resources, resource)
+                            fitDrawable(textWithMarkup, bd)
+                            text.setSpan(ImageSpan(bd), text.getSpanStart(span), text.getSpanEnd(span), text.getSpanFlags(span))
                         }
                     })
             }
@@ -106,14 +115,12 @@ object ImageLoader {
         return ret
     }
 
-    fun fitDrawable(view: View, drawable: Drawable) {
+    fun fitDrawable(view: View, width: Int, height: Int): Pair<Int, Int> {
+        var newWidth = width
+        var newHeight = height
+
         // Get the display metrics to be able to rescale the image if needed
         val metrics = view.context.resources.displayMetrics
-
-        // Gather drawable information
-        // Scale the height by the scaledDensity to get original image size
-        val drawableHeight = drawable.intrinsicHeight.toFloat() * metrics.scaledDensity
-        val drawableWidth = drawable.intrinsicWidth.toFloat() * metrics.scaledDensity
 
         // Use either a fixed size or a scaled size according to user preferences
         var fixedSize = -1
@@ -121,22 +128,38 @@ object ImageLoader {
             fixedSize = AppPreferences.setImageFixedWidth(view.context)
         }
 
+        // Before image loading view.width might not be initialized
+        // So we take a default maximum value that will be reduced
+        var maxWidth = view.width
+        if(maxWidth == 0)
+        {
+            maxWidth = metrics.widthPixels
+        }
+
+        // If we are using a fixedSize
         if (fixedSize > 0) {
             // Keep aspect ratio when using fixed size
-            val ratio = drawableHeight / drawableWidth
-            drawable.setBounds(0, 0, fixedSize, (fixedSize * ratio).toInt())
-        } else {
-            // Rescale the drawable if it is larger that the current view width
-            if (drawableWidth > view.width) {
-                //Compute image ratio
-                val ratio = drawableHeight / drawableWidth
-                // Ensure that the images have a minimum size
-                val width = Math.max(view.width, 256).toFloat()
+            val ratio = height.toFloat() / width.toFloat()
+            newWidth = fixedSize
+            newHeight = (fixedSize * ratio).toInt()
+        // Otherwise if we are using rescaling and the image is wider that the max width
+        } else if (width > maxWidth) {
+            //Compute image ratio
+            val ratio = height.toFloat() / width.toFloat()
+            // Ensure that the images have a minimum size
+            val width = Math.max(maxWidth, 256).toFloat()
 
-                drawable.setBounds(0, 0, width.toInt(), (width * ratio).toInt())
-            } else {
-                drawable.setBounds(0, 0, drawableWidth.toInt(), drawableHeight.toInt())
-            }
+            newWidth = width.toInt()
+            newHeight = (width * ratio).toInt()
         }
+
+        return Pair(newWidth, newHeight)
+    }
+
+    fun fitDrawable(view: View, drawable: BitmapDrawable) {
+        // Compute the new size of the drawable
+        val newSize = fitDrawable(view, drawable.bitmap.width, drawable.bitmap.height)
+        // Set the bounds to match the new size
+        drawable.setBounds(0, 0, newSize.first, newSize.second)
     }
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -161,3 +161,10 @@ def orgJavaLocation() {
         return "com.orgzly:org-java:$org_java_version"
     }
 }
+
+apply plugin: 'kotlin-kapt'
+
+dependencies {
+    implementation 'com.github.bumptech.glide:glide:4.8.0'
+    kapt 'com.github.bumptech.glide:compiler:4.8.0'
+}
--- a/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
+++ b/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
@@ -542,6 +542,37 @@ public class AppPreferences {
     }
 
     /*
+     * Allow inlining images
+     */
+    public static boolean displayInlineImages(Context context) {
+        return getDefaultSharedPreferences(context).getBoolean(
+                context.getResources().getString(R.string.pref_key_display_inline_images),
+                context.getResources().getBoolean(R.bool.pref_default_display_inline_images));
+    }
+
+    public static void displayInlineImages(Context context, boolean value) {
+        String key = context.getResources().getString(R.string.pref_key_display_inline_images);
+        getDefaultSharedPreferences(context).edit().putBoolean(key, value).apply();
+    }
+
+    public static boolean enableImageScaling(Context context) {
+        return getDefaultSharedPreferences(context).getBoolean(
+                context.getResources().getString(R.string.pref_key_enable_image_scaling),
+                context.getResources().getBoolean(R.bool.pref_default_enable_image_scaling));
+    }
+
+    public static void enableImageScaling(Context context, boolean value) {
+        String key = context.getResources().getString(R.string.pref_key_enable_image_scaling);
+        getDefaultSharedPreferences(context).edit().putBoolean(key, value).apply();
+    }
+
+    public static int setImageFixedWidth(Context context) {
+        return Integer.valueOf(getDefaultSharedPreferences(context).getString(
+                context.getResources().getString(R.string.pref_key_set_image_fixed_width),
+                context.getResources().getString(R.string.pref_default_set_image_fixed_width)));
+    }
+
+    /*
      * Note's metadata visibility
      */
 
--- a/app/src/main/res/values/prefs_keys.xml
+++ b/app/src/main/res/values/prefs_keys.xml
@@ -354,6 +354,15 @@
     <string name="pref_key_set_last_repeat_on_time_shift" translatable="false">pref_key_set_last_repeat_on_time_shift</string>
     <bool name="pref_default_set_last_repeat_on_time_shift" translatable="false">true</bool>
 
+    <string name="pref_key_display_inline_images" translatable="false">pref_key_display_inline_images</string>
+    <bool name="pref_default_display_inline_images" translatable="false">true</bool>
+
+    <string name="pref_key_enable_image_scaling" translatable="false">pref_key_enable_image_scaling</string>
+    <bool name="pref_default_enable_image_scaling" translatable="false">true</bool>
+
+    <string name="pref_key_set_image_fixed_width" translatable="false">pref_key_set_image_fixed_width</string>
+    <string name="pref_default_set_image_fixed_width" translatable="false">128</string>
+
     <!-- Separate notes with an empty line -->
     <string name="pref_key_separate_notes_with_new_line" translatable="false">pref_key_separate_notes_with_new_line</string>
     <string name="pref_default_separate_notes_with_new_line" translatable="false">@string/pref_value_separate_notes_with_new_line_multi_line_notes_only</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -538,4 +538,11 @@
 
     <string name="external_file_no_app_found">No application found to open this file</string>
 
+    <string name="display_inline_images">Display inline images</string>
+    <string name="display_inline_images_summary">Display inline images in notes</string>
+
+    <string name="enable_image_scaling">Enable image scaling</string>
+    <string name="enable_image_scaling_summary">Use predefined sizes or enable image scaling</string>
+
+    <string name="pref_title_set_image_fixed_width">Set inlined image fixed width (in pixels)</string>
 </resources>
--- a/app/src/main/res/xml/prefs_screen_notebooks.xml
+++ b/app/src/main/res/xml/prefs_screen_notebooks.xml
@@ -2,6 +2,7 @@
 
 <PreferenceScreen
     xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:title="@string/pref_title_notebooks">
 
     <PreferenceCategory android:title="@string/notebooks">
@@ -148,6 +149,27 @@
             android:summary="@string/set_last_repeat_on_time_shift_summary"
             android:defaultValue="@bool/pref_default_set_last_repeat_on_time_shift"/>
 
+        <SwitchPreference
+            android:key="@string/pref_key_display_inline_images"
+            android:title="@string/display_inline_images"
+            android:summary="@string/display_inline_images_summary"
+            android:defaultValue="@bool/pref_default_display_inline_images"/>
+
+        <SwitchPreference
+            android:key="@string/pref_key_enable_image_scaling"
+            android:title="@string/enable_image_scaling"
+            android:summary="@string/enable_image_scaling_summary"
+            android:defaultValue="@bool/pref_default_enable_image_scaling"
+            android:dependency="@string/pref_key_display_inline_images"
+            android:disableDependentsState="true"/>
+
+        <com.orgzly.android.prefs.IntegerPreference
+            android:key="@string/pref_key_set_image_fixed_width"
+            android:title="@string/pref_title_set_image_fixed_width"
+            android:inputType="number"
+            android:defaultValue="@string/pref_default_set_image_fixed_width"
+            app:min="1"
+            android:dependency="@string/pref_key_enable_image_scaling"/>
     </PreferenceCategory>
 
     <PreferenceCategory android:title="@string/prefs_title_new_note">


nevenz:
> I also added an ImageSpan to display images in notes

Cool!

> rescaled to fit width, if larger than the window width

What's the expected size of the displayed image when it's too wide? I'm seeing half a screen width in portrait mode and the same size in landscape. Should it depend on orientation?

> file:Download/test2.png

I don't think this is displayed in Org mode. Only images in square bracket links. I think we should do the same.

> Right now, named links are not replaced with an ImageSpan.

Yeah, we'll want these displayed too.

> Furthemore regarding to the application performance, maybe adding a setting to toggle off image display could be a good idea. What do you think about this ?

Sounds good. Perhaps even keeping the feature off by default, though I'm not sure about that.

I'll leave few comments in the code too.

Thanks!

aancel:
Ok, I've updated the code a bit, but there are still missing points:

> What's the expected size of the displayed image when it's too wide? I'm seeing half a screen width in portrait mode and the same size in landscape. Should it depend on orientation?

Here I am not very happy with the use of `currentActivity.window.decorView.width`. This variable indicates the current width of the window (e.g. 1920 or 1080 for a 1080p device).
Normally, if the image is smaller than the current width (which should work in both portrait and landscape mode), it doesn't get resized (which can end up with very small pictures ...), but if it is larger than the current width, it is resized to match the max width, so it is viewable.
I did not encouter issues with my test images however ... I'll give it another try.
If you have ideas on a better way to get a max width, I'm open to suggestions !

> > file:Download/test2.png
> 
> I don't think this is displayed in Org mode. Only images in square bracket links. I think we should do the same.

I usually don't display my images in org-mode, so it did not bother me to have it displayed, but I can disable that if you prefer.

> 
> > Right now, named links are not replaced with an ImageSpan.
> 
> Yeah, we'll want these displayed too.

I haven't fixed this yet. It seems to behave differently than the unnamed links, and I was not able to find why as of now.

> 
> > Furthemore regarding to the application performance, maybe adding a setting to toggle off image display could be a good idea. What do you think about this ?
> 
> Sounds good. Perhaps even keeping the feature off by default, though I'm not sure about that.

I've added a boolean in the Config object, this should make it easier to set off.

> Few more things...
> 
> Displayed images are now clickable too. Do we want that behavior? I don't really have a preference.
> 

I think this is a good idea to keep this, in case the user wants to display it in an image viewer for zooming (for small images) or modifications, in the same way you can open external links in org-mode.

> Images can now be included and displayed in note title too. I don't think we should support that, even though it seems kinda cool. It's messing up the layout which could be tricky to fix. And even when fixed, I think notes would be harder to read with variable height titles (multi-line or not). Then again, Org mode support that.

This should be fixed with the additional config parameter I added. I also took the liberty to remove it from the reminders, as It would probably also mess with the layout in narrow spaces.

> Links like `[[foo]]` are now linkified, leaving square brackets visible. Is there a clean and fast way not to do that unless it's a path to existing image? Org mode would look for a heading name, which is not supported in Orgzly (yet). Probably not a big deal, just a bit confusing (to be at least).

I've modified the function generating a ClickableSpan, so it does not create a link when the link is not valid for a file. It kinds of implement indicating that the file is invalid, as we discussed in the previous PR.
However this causes a `File` instance and `file.exists()` check to be done a bit earlier, which is something that you wanted to avoid in a previous comment. So if you prefer the previous way, I'll revert back to this.

> Code style needs some fixing. Space after `if` is missing in few places. And for consistency, `else` would have to be moved to the same line as `}`. Though I'm just nitpicking now.

The code style should be fixed hopefully !

app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
```diff
@@ -347,66 +400,100 @@ object OrgFormatter {
...
-                                    }
-                                } else {
-                                    currentActivity.showSnackbar(context.getString(R.string.file_does_not_exist, file.absolutePath))
-                                }
+                // Ensure the file exists
+                val file = File(Environment.getExternalStorageDirectory(), path)
+                if (file.exists()
+                        && (file.extension.toLowerCase() == "png"
+                                || file.extension.toLowerCase() == "jpg"
+                                || file.extension.toLowerCase() == "bmp")
```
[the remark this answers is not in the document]
aancel: I've update this code, and the supported extensions via the ImageUtils class.
I did not find a portable way to know the supported image formats, so I used the ones described in the android documentation: https://developer.android.com/guide/topics/media/media-formats

app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
```diff
@@ -154,9 +151,31 @@ object OrgFormatter {
...
+                    var cs: ClickableSpan? = null
+
+                    if(linkify) {
+                        // Ensure we have a path component
+                        if (s.size == 2) {
+                            cs = configureFileLink(ssb, s[1], m.start(), m.end())
+                        } else {
+                            cs = configureFileLink(ssb, link, m.start(), m.end())
+                        }
+                    }
```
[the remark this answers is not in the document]
aancel: If I remember well, the `== 1` is because when splitting a string without the specified delimiter, the method will return an Array of size 1 with the original string, so it should handle the case of not having a `file:` scheme.
Indeed I'll try to simplify this !

nevenz:
> I did not encouter issues with my test images however ... I'll give it another try.

I've been using 2560x1600 JPEG and seeing:

- decorWidth: 1440
- drawableHeight: 457
- drawableWidth: 731

for whatever reason.

> If you have ideas on a better way to get a max width, I'm open to suggestions !

It's tricky because of possible indentation too. Higher-level notes will have less space available. Also, there could be text before and after the image on the same line, though that will wrap.

This is what I'm seeing:

orgzly-photos

Which looks fine, but it's unexpected. I thought image would take more space. I haven't tested it with different density devices.

>> > Right now, named links are not replaced with an ImageSpan.
>> 
>> Yeah, we'll want these displayed too.
>
> I haven't fixed this yet. It seems to behave differently than the unnamed links, and I was not able to find why as of now.

Perhaps adding some tests to `OrgFormatterTest` might help. I don't see anything wrong at first glance.

> However this causes a `File` instance and `file.exists()` check to be done a bit earlier, which is something that you wanted to avoid in a previous comment. So if you prefer the previous way, I'll revert back to this.

I'm more concerned about the speed of loading images now. `Drawable.createFromStream` is very slow. When you scroll in a notebook with a lot of images (like the one in the screenshot), it's *very* choppy. And that's with only two images being loaded over and over again. Perhaps they can be loaded in background at least. Maybe even caching drawables. We'll have to look into how that's done.

BTW, I noticed `[[Download/logo.jpg/]]` is creating a link leaving square brackets visible. `Download/logo.jpg` does exist (and it's opened on tap, which is also weird), but this is a link to `Download/logo.jpg/`, so it shouldn't be linkified. This is another good case to check in tests.

nevenz:
> However I don't think that it really makes sense to further reduce the size of the images as the note level increases, in the end you would have really small images forced to fit in the view, and images that would disappear because they would be outside of the visible area.
So I think it might be better to have a reasonable resizing and having the user being conscient about it.

Yes, I agree. It's just that I don't understand why this huge image doesn't get to fill the whole width. So I'm worried that there could be issues on some other devices. But if we check few more different sizes and densities and it looks OK, I'm fine with that.

> Regarding to the loading speed of the images, there seem to be some LruCache available on Android, see https://developer.android.com/topic/performance/graphics/cache-bitmap.
I think that it might be a good idea to use this.
There are also third party libraries that handle image loading in the background, like Picasso or Glide. What do you think about using one of those ? (I'm asking because you might want to avoid including external dependencies).
Combined with the LruCache, I think it can help.

Notebook in the screenshot I posted might not be a good example of a typical usage. Most likely the images will be all different. I can easily see users having bunch of them in many notes. So, caching might not be as important as loading images in the background. Not that we shouldn't try to use cache if it's not too hard.

If Glide can help in any way, great. We shouldn't write more code then we need to, I don't mind external dependency unless it's for something trivial.

aancel:
I've pushed another update, implementing image asynchronous loading and caching, which should hopefully improve performance. This was harder that I thought at the beginning ! Especially with the use of ImageSpan (It seems easier with classical views)
I've also added preferences to enable the display of images and specify whether the user wants the images to scale or to have a fixed size.

I ended up not using glide, as it was not that easy to use, and they seem to be changing APIs right now, with the deprecation of some objects. I used an AsyncTask and the LruCache.
I've also passed the TextViewWithMarkup to OrgFormatter. I'm not really a fan of this, but it gives more flexibility on the scaling of images.

One issue that I have now is with the preferences. When toggling the new ones I added, the settings don't get applied until a restart of the application. If you have any clue of what I might have done wrong, please tell me, so I can fix this.

I also reworked the code with the `scheme` and `file:` handling and splitting.

I've also tested the app with images similar in size of the big ones you use, loading and scaling is much better for me now.

nevenz:
I don't see any images now, just linkified `[[Download/logo.jpg]]`. Image is displayed when clicked on. I've tried toggling the option (including leaving the app), but also just replacing AppPreferences call to `true`.

Wouldn't this also have issues due to view recycling? And also potentially start bunch of `AsyncTask`s when scrolling fast?

I don't think this'll be that easy to get it right. We should definitely check how it's being done in some apps out there. It's a common thing to do, though I agree, perhaps not so much with `ImageSpan`. But it should be similar.

> I've also passed the TextViewWithMarkup to OrgFormatter. I'm not really a fan of this, but it gives more flexibility on the scaling of images.

I think it's OK, but then I'd remove `OrgFormatter.parse` from `TextViewWithMarkup.setText` and do it the other way around - call new `OrgFormatter.setText(TextViewWithMarkup, String)` on the view.

BTW, I don't see anything wrong with the preference - when I install app for the first time, it's set to `true`.

aancel:
> I don't see any images now, just linkified [[Download/logo.jpg]]. Image is displayed when clicked on. I've tried toggling the option (including leaving the app), but also just replacing AppPreferences call to true.

Do you still use the same platforms as in the previous PRs ? Do you have the same issue on all the platforms ?
I've tested with Pixel 2 API 23 and Pixel 2 CL API 27, and it seems to be working fine. I'm having some issues with the Nexus 5 API 16 however. I'll start working on this trying to see if the issue behind is related to what you encounter.

> Wouldn't this also have issues due to view recycling? And also potentially start bunch of AsyncTasks when scrolling fast?

While scrolling fast I didn't encounter any issue on my side.

> I don't think this'll be that easy to get it right. We should definitely check how it's being done in some apps out there. It's a common thing to do, though I agree, perhaps not so much with ImageSpan. But it should be similar.

I'll look for similar stuff, but yeah it's less common with ImageSpan that classical views.

> I think it's OK, but then I'd remove OrgFormatter.parse from TextViewWithMarkup.setText and do it the other way around - call new OrgFormatter.setText(TextViewWithMarkup, String) on the view.

Ok, I'll go this way, once I sorted out the loading issues.

Also regarding your previous comment:
> BTW, I noticed [[Download/logo.jpg/]] is creating a link leaving square brackets visible. Download/logo.jpg does exist (and it's opened on tap, which is also weird), but this is a link to Download/logo.jpg/, so it shouldn't be linkified. This is another good case to check in tests.

I can also reproduce this, but it seems that `File` considers that `Download/logo.jpg/` is a file, when it should be considered as a folder.

nevenz:
> Do you still use the same platforms as in the previous PRs ? Do you have the same issue on all the platforms ?

Yes, I also tried to checkout the previous commit and I do see the images there. Then going back to latest one, I don't.

> I'm having some issues with the Nexus 5 API 16 however.

What are you seeing there? I haven't tried API 16.

> While scrolling fast I didn't encounter any issue on my side.

I suspect there could be updates to the wrong views, but we'll see.

> I can also reproduce this, but it seems that File considers that Download/logo.jpg/ is a file, when it should be considered as a folder.

But this shouldn't pass the check for extension though?

aancel:
> Yes, I also tried to checkout the previous commit and I do see the images there. Then going back to latest one, I don't.

It's really very strange, because I've tested it on a fresh image with a Nexus 6P API 25 and it works as expected... I don't get where the issue is.
Could you try to put some log in the AsyncImageLoader.java class, notably in the doInBackground function to see if the app goes here and whether it loads image from the cache or not ?
It will at least tell us if the threads are launched and maybe if it a rendering issue. If not, then the issue should probably be in the tests of OrgFormatter.kt.

> What are you seeing there? I haven't tried API 16.

The issue here was with a bad test that broke the image display when the image is at the very end of the note and that the `]]` closes the note.

> I suspect there could be updates to the wrong views, but we'll see.

There is an additional problem with views: When the storage permission is not granted at the start of the app, then launch the app, try to click on a link to make the permission snackbar appear, grant the permission, then back to the app, the image are not immediately refreshed. I suspect that the view are not refreshed when resuming the app.

> But this shouldn't pass the check for extension though?

The extension check is only performed when we want to display an image with an ImageSpan. 
Classical link (via ClickableSpan) can be any file like a pdf, txt or an image and are thus linkified.

I don't know if I will have much time this evening, but I'll do my best to sort all those issues !

nevenz:
>> What are you seeing there? I haven't tried API 16.
> The issue here was with a bad test that broke the image display when the image is at the very end of the note and that the ]] closes the note.

Ah, it looks like this is it. All my notes (well, only 3 unique, repeated bunch of times) are like that. Images are displayed when I add a space after `]]`.

Few more things I noticed...

Images stop being displayed when I:
- Enter Settings
- Press Back
- Press Back (to return to notebook list)
- Enter the same notebook with images again

When scrolling up slowly, there is sudden jump as note comes into view, I assume due to image being inserted and note now occupying more space.

With scaling option on, images occupy maximum width, even if there is some text after them, which then wraps. Might be OK, I'm not sure.

With scaling option off, aspect ratio doesn't seem to be kept.

When toggling above option, only images from first few notes are displayed. When I keep scrolling down, there are no images. When I go back to the top, few images previously displayed are gone.

aancel:
> Ah, it looks like this is it. All my notes (well, only 3 unique, repeated bunch of times) are like that. Images are displayed when I add a space after ]].

This should be fixed in the latest commit !

> With scaling option off, aspect ratio doesn't seem to be kept.

I also updated this, so only the width is fixed. The height is computed to keep the aspect ratio.

> Images stop being displayed when I:
>
>    Enter Settings
>    Press Back
>    Press Back (to return to notebook list)
>    Enter the same notebook with images again

I have a similar issue, when I toggle one of the settings ...
Even worse, when I toggle one of the setting, and go back, then back to the note (There the images are not displayed any longer). If I go back to the notebook list, I cannot go back into the notebook I exited from ... I have to relaunch the application to access it again. That's why I think there might have been an issue with the settings, or maybe it would be related to the views not being refreshed.
Would you have an idea where to look for ? Somewhere in the onResume of MainActivity maybe ?

> When scrolling up slowly, there is sudden jump as note comes into view, I assume due to image being inserted and note now occupying more space.

Yes, I agree it is a bit disturbing, and I don't really know how to compensate for that.

> With scaling option on, images occupy maximum width, even if there is some text after them, which then wraps. Might be OK, I'm not sure.

This is indeed debatable, and that's also why I put a scaling parameter to give the choice to the user.
I suppose that the people who want full width images won't wrap text around.
(Feedback from users about use-cases would really be valuable here)

nevenz:
> I also updated this, so only the width is fixed. The height is computed to keep the aspect ratio.

Cool.

In both cases (scaling or not) - I don't think image should be scaled *up*. Or at least there should be another option to control that?

Also, current hardcoded `256` fixed image size should probably be configurable too.

> Would you have an idea where to look for ? Somewhere in the onResume of MainActivity maybe ?

Current (tracked) activity is lost after returning from Settings:

```
10-10 09:53:48.600   665   665 D com.orgzly.android.ui.CommonActivityLifecycleCallbacks: 2#main: onActivityResumed: WIP com.orgzly.android.ui.MainActivity@8cf807
10-10 09:53:48.600   665   665 D com.orgzly.android.ui.CommonActivityLifecycleCallbacks: 2#main: registerActivity: WIP com.orgzly.android.ui.MainActivity@8cf807
10-10 09:53:48.850   665   665 D com.orgzly.android.ui.CommonActivityLifecycleCallbacks: 2#main: onActivityStopped: WIP com.orgzly.android.ui.settings.SettingsActivity@303ac9c
10-10 09:53:48.850   665   665 D com.orgzly.android.ui.CommonActivityLifecycleCallbacks: 2#main: registerActivity: WIP null
```

`registerActivity` should only be tracking `CommonActivity`. A simple:

```
    @Override
    public void onActivityResumed(Activity activity) {
        if (activity instanceof CommonActivity) {
            App.setCurrentActivity((CommonActivity) activity);
        }
    }

    @Override
    public void onActivityPaused(Activity activity) {
        if (activity instanceof CommonActivity) {
            App.setCurrentActivity(null);
        }
    }
```

should be enough, removing everything from other methods.

I don't know if this is also the reason for the issue you seeing - not being able to return to the notebook.

> Yes, I agree it is a bit disturbing, and I don't really know how to compensate for that.

Without knowing image dimension in advance, it's probably not possible. User experience could be improved though:

- Move cache check out from `AsyncImageLoader`. When `Bitmap` is found in cache, display the image immediately. Only helps if image is cached of course. But I think this is the Right Thing to do anyway - `AsyncImageLoader` should only be needed if we have to read the stream. 
- Display a placeholder image before calling `AsyncImageLoader`. This should decrease the difference in height when the real image is displayed.
- Animate item expansion

nevenz:
I've fixed [snackbar issue](https://github.com/orgzly/orgzly-android/pull/382#issuecomment-425973657) in master, hence the conflict, though it should be easy to resolve.

I've also created [OrgFormatterLinkTest](https://github.com/orgzly/orgzly-android/blob/master/app/src/androidTest/java/com/orgzly/android/util/OrgFormatterLinkTest.kt) so we can list all the cases - should the link be parsed or not, clickable or not etc.

This test is failing right now, there are two fixes needed in `OrgFormatter`, but I didn't want to edit it more. We should have split this PR - one for rework, one for images but oh well.

aancel:
Ok, thanks for the update !
Should I rebase on or merge the latest master to continue working on this ?

Yes indeed I agree for the split.

nevenz:
> Should I rebase on or merge the latest master to continue working on this ?

I prefer rebase, but whatever is easier for you. Thanks!

aancel:
> In both cases (scaling or not) - I don't think image should be scaled _up_. Or at least there should be another option to control that?

With scaling on, images should never be scaled up. They are either kept to their original size if they fit within the view width or scaled down to fit the current view width.
With scaling off, the images might be scaled up if the user desires indeed.
But maybe, here you are referring to the multiplication of the drawable size by `metrics.scaledDensity` ? This was done to recover the original image size, as apparently android seem to be scaling down the images with the density.
This indeed could be configurable.
Did you mean anything else ?

> Also, current hardcoded `256` fixed image size should probably be configurable too.

I agree, this was mainly as a first test, I'll add a preference to set this.

> 
> Current (tracked) activity is lost after returning from Settings:
> 
> ```
> 10-10 09:53:48.600   665   665 D com.orgzly.android.ui.CommonActivityLifecycleCallbacks: 2#main: onActivityResumed: WIP com.orgzly.android.ui.MainActivity@8cf807
> 10-10 09:53:48.600   665   665 D com.orgzly.android.ui.CommonActivityLifecycleCallbacks: 2#main: registerActivity: WIP com.orgzly.android.ui.MainActivity@8cf807
> 10-10 09:53:48.850   665   665 D com.orgzly.android.ui.CommonActivityLifecycleCallbacks: 2#main: onActivityStopped: WIP com.orgzly.android.ui.settings.SettingsActivity@303ac9c
> 10-10 09:53:48.850   665   665 D com.orgzly.android.ui.CommonActivityLifecycleCallbacks: 2#main: registerActivity: WIP null
> ```
> 
> `registerActivity` should only be tracking `CommonActivity`. A simple:
> 
> ```
>     @Override
>     public void onActivityResumed(Activity activity) {
>         if (activity instanceof CommonActivity) {
>             App.setCurrentActivity((CommonActivity) activity);
>         }
>     }
> 
>     @Override
>     public void onActivityPaused(Activity activity) {
>         if (activity instanceof CommonActivity) {
>             App.setCurrentActivity(null);
>         }
>     }
> ```
> 
> should be enough, removing everything from other methods.

Indeed, I remove the else part of the registerActivity test to implement what you propose

> I don't know if this is also the reason for the issue you seeing - not being able to return to the notebook.

This doesn't seem to be fixed. I'll investigate this.

> > Yes, I agree it is a bit disturbing, and I don't really know how to compensate for that.
> 
> Without knowing image dimension in advance, it's probably not possible. User experience could be improved though:
> 
>     * Move cache check out from `AsyncImageLoader`. When `Bitmap` is found in cache, display the image immediately. Only helps if image is cached of course. But I think this is the Right Thing to do anyway - `AsyncImageLoader` should only be needed if we have to read the stream.
> 
>     * Display a placeholder image before calling `AsyncImageLoader`. This should decrease the difference in height when the real image is displayed.
> 
>     * Animate item expansion

Great, I'll see what I can do on this side, although I'm just not sure that item animation is possible.

> I prefer rebase, but whatever is easier for you. Thanks!

I can go for a rebase. I prefer asking because some people don't like forced pushes !

nevenz:
> With scaling on, images should never be scaled up. 
> Did you mean anything else ?

No, you're right. It due to different issue I encountered.

I pushed a new tiny image to the emulator by overwriting the old one, left the list (or app), returned to it and went back to the book. I saw the same image, not the new one. It's due to caching by name probably and app didn't terminate.

I don't think we need to keep the cache during entire app run. It's enough to create it when user enters the notebook (or search results). `NoteListFragment` might be a good place to do it as it's used by both notebook and search results.

But even then, we'll need a way to invalidate the entires. And/or store them by something other then just file name. I'd much rather have cache misses here, then stale data. Cache is much less important now with async loading. I don't consider this a blocker for this PR though. Just moving the cache from `App` will do for now. (I'd also suggest wrapping that whole code in a new class.)

>> I don't know if this is also the reason for the issue you seeing - not being able to return to the notebook.
> This doesn't seem to be fixed. I'll investigate this.

Is it #348 or similar perhaps? That would be helpful, as I wasn't able to reproduce that one.

> Great, I'll see what I can do on this side, although I'm just not sure that item animation is possible.

There's probably a way (worst case generating two item views and animating the replacement, or something like that). But definitely not a blocker for this PR, we can do that later.

aancel:
Okay, rebase has been done on the latest commit. I've encountered 2 merge conflicts, I had a look at your commits, I hope to have done the things correctly with this rebase.

Regarding to the permission snackbar, I am now having the same issue as I had before.
To repoduce it on my side:
- Remove the storage permission
- Go in the app and click on a link
- On my side, the snackbar pops up, then immediatly closes, so permission can't be configured

> I pushed a new tiny image to the emulator by overwriting the old one, left the list (or app), returned to it and went back to the book. I saw the same image, not the new one. It's due to caching by name probably and app didn't terminate.

> I don't think we need to keep the cache during entire app run. It's enough to create it when user enters the notebook (or search results). NoteListFragment might be a good place to do it as it's used by both notebook and search results.

> But even then, we'll need a way to invalidate the entires. And/or store them by something other then just file name. I'd much rather have cache misses here, then stale data. Cache is much less important now with async loading. I don't consider this a blocker for this PR though. Just moving the cache from App will do for now. (I'd also suggest wrapping that whole code in a new class.)

Regarding this kind of change of the image, maybe using a hash as the storage key instead of the path could solve the issue ? Adding the overhead of hash computation, but it would then be a reliable cache

aancel:
> Is it #348 or similar perhaps? That would be helpful, as I wasn't able to reproduce that one.

It indeed seems to be the same thing that is happening to me. The `Force UTF-8` boolean also triggers the issue for me. It would be great to also solve this issue at the same time.

aancel:
Regarding the broken UI actions, I might have found something, but not so sure. I report it here so maybe it will help.

I focused on opening a notebook which works in the normal case, but not in the buggy case (meaning having toggled a button in the preference).

I used the debugger and located an area that should be the cause of not opening the notebook.
I used the debugger on this line:
https://github.com/orgzly/orgzly-android/blob/8c8e7be2b2d7a3085a287ac6b680bb97a82c0913/app/src/main/java/com/orgzly/android/ui/MainActivity.java#L1162

In the normal case, and going trough all the functions I end up in the LocalBroadcastManager on line 255:
```
            ArrayList<ReceiverRecord> entries = mActions.get(intent.getAction());
            if (entries != null) {
                if (debug) Log.v(TAG, "Action list: " + entries);
```
Here `entries` is not null and thus proceeds into the code.
But in the buggy case, `entries` is null and it never goes in the if scope and the function returns false.

nevenz:
> Regarding to the permission snackbar, I am now having the same issue as I had before.
> On my side, the snackbar pops up, then immediatly closes, so permission can't be configured

Ah yes, fix is only for snackbar called from `OrgFormatter`, which doesn't even have an action. Cheap solution would be to wrap everything in `Handler().postDelayed`, so that `onClick` finishes first. I'll look into it again.

> Regarding this kind of change of the image, maybe using a hash as the storage key instead of the path could solve the issue ? Adding the overhead of hash computation, but it would then be a reliable cache

I don't know if it would be fast enough. You'd have to open and read entire stream, calculating the hash. Not to mention you'd have to do open and read the stream again if you get a miss, unless you kept the content the first time. Though that would be done asynchronous. It's worth testing I guess.

> It indeed seems to be the same thing that is happening to me. The Force UTF-8 boolean also triggers the issue for me. It would be great to also solve this issue at the same time.

It might be something to do with activities being restarted on preference change. Try disabling `SharedPreferences.OnSharedPreferenceChangeListener` in `CommonActivity`.

aancel:
I've taken screenshots of the UI while debugging this gives a hint at the values when the issue occurs:
The normal case:
orgzly normal opening notebook

The case of notebook not opening:
orgzly bug opening notebook

The action associated to the intent does not seem to be found in the bad case.

In the ok case, mActions contains:
- "com.orgzly.intent.action.UPDATING_NOTES_STARTED" -> " size = 1"
- "com.orgzly.intent.action.DB_UPGRADE_STARTED" -> " size = 1"
- "com.orgzly.intent.action.OPEN_QUERY" -> " size = 1"
- "com.orgzly.intent.action.OPEN_BOOK" -> " size = 1"
- "com.orgzly.intent.action.SHOW_SNACKBAR" -> " size = 1"
- "com.orgzly.intent.action.BOOK_IMPORTED" -> " size = 1"
- "com.orgzly.intent.action.OPEN_BOOKS" -> " size = 1"
- "com.orgzly.intent.action.DB_UPGRADE_ENDED" -> " size = 1"
- "com.orgzly.intent.action.OPEN_SETTINGS" -> " size = 1"
- "com.orgzly.intent.action.OPEN_NOTE" -> " size = 1"
- "com.orgzly.intent.action.OPEN_QUERIES" -> " size = 1"
- "com.orgzly.intent.action.DB_CLEARED" -> " size = 1"
- "com.orgzly.intent.action.SYNC_STATUS" -> " size = 1"
- "com.orgzly.intent.action.UPDATING_NOTES_ENDED" -> " size = 1"

In the bad case, mActions contains:
- "com.orgzly.intent.action.UPDATING_NOTES_STARTED" -> " size = 1"
- "com.orgzly.intent.action.DB_UPGRADE_STARTED" -> " size = 1"
- "com.orgzly.intent.action.SHOW_SNACKBAR" -> " size = 1"
- "com.orgzly.intent.action.BOOK_IMPORTED" -> " size = 1"
- "com.orgzly.intent.action.DB_UPGRADE_ENDED" -> " size = 1"
- "com.orgzly.intent.action.DB_CLEARED" -> " size = 1"
- "com.orgzly.intent.action.SYNC_STATUS" -> " size = 1"
- "com.orgzly.intent.action.UPDATING_NOTES_ENDED" -> " size = 1"

aancel:
Latest small update refactors the code a bit, ensure that we only register the lifecycle of CommonActivity and adds a setting to update the size of fixed-width images.

I'll have a look at your comment on missing actions. I'll see if it solves the issue for me.
This huge PR might not be the right place to put this code if it solves the issue. What do you think about this ?

nevenz:
> This huge PR might not be the right place to put this code if it solves the issue. What do you think about this ?

I agree, It should definitely be a new PR and we can move any further conversation to #348.

nevenz:
> Regarding to the permission snackbar, I am now having the same issue as I had before.

It should be (really) fixed now.

aancel:
> It should be (really) fixed now.

Yes, I confirm it is fixed ! (I rebased the current PR code to be up to date)

aancel:
I've pushed a new update:
- A check has been added for files ending with '/'. If this is a directory, we linkify it, otherwise it won't be linkified.
- A transparent placeholder image has been put in place while the image is loading in background. I set the placeholder size to the fixed size specified by the user (It works well without scaling, and should improve the situation with scaling)
- AppPreferences have been updated, so that the Display inline images toggle enables or disables the 2 others
- I think that images could still be displayed in titles and reminders as I switched to preferences. It should not be the case any longer.

A few other remarks/ideas:
- What do you think about using file size to invalidate images in the cache, i.e. using file.length()? 
I've timed its use and I was always getting 0 ms even for 2560 * 1600 images. It could be a rough estimate to discard cached images.
- About the cache too, it seems that the chosen cache amount is really small. I could only fit 2 to 3 images in it. Should I increase its size ? I'll be moving it to the Notebook too, do you think it should be there per instance or as a class member ?

app/src/main/java/com/orgzly/android/ui/settings/SettingsFragment.kt
```diff
@@ -277,6 +279,17 @@ class SettingsFragment : PreferenceFragment(), SharedPreferences.OnSharedPrefere
...
+        val displayInlineImages = findPreference(getString(R.string.pref_key_display_inline_images))
+
+        if (displayInlineImages != null) {
+            val displayInlineImagesEnables = (displayInlineImages as TwoStatePreference).isChecked;
+
+            findPreference(getString(R.string.pref_key_enable_image_scaling))?.isEnabled = displayInlineImagesEnables
+            findPreference(getString(R.string.pref_key_set_image_fixed_width)).isEnabled = displayInlineImagesEnables
+        }
+    }
+
```
nevenz: This should be done by [dependency attributes](https://developer.android.com/reference/android/preference/Preference.html#attr_android:dependency).

It's done in code for reminders because those preferences depend on *two* other (scheduled and deadline).

I think `pref_key_enable_image_scaling` should depend on `pref_key_display_inline_images` and `pref_key_set_image_fixed_width` should depend on `pref_key_enable_image_scaling`. `pref_key_enable_image_scaling` would be "Use fixed-size image width".

nevenz:
> What do you think about using file size to invalidate images in the cache, i.e. using file.length()?
I've timed its use and I was always getting 0 ms even for 2560 * 1600 images. It could be a rough estimate to discard cached images.

Sure. Modification time could be included too.

> About the cache too, it seems that the chosen cache amount is really small. I could only fit 2 to 3 images in it. Should I increase its size ?

I think that's fine for now.

> I'll be moving it to the Notebook too, do you think it should be there per instance or as a class member ?

I'm not sure I understand where did you plan to move it? Note that there could be search results with notes from different notebooks displaying images too.

BTW, I'm getting:

image

with both preferences enabled, when I return from Settings and start scrolling down.

nevenz:
Another issue - Trying a notebook with 20 notes having different large image each (taken from https://unsplash.com/search/photos/large), app and services on devices are killed due to memory issues. It's enough to just enter the notebook, although it happens faster when you scroll down. This is on Pixel 2 XL API 27 emulator with default options.

nevenz:
I've removed this from v1.6.10 milestone until we iron it out.

As I want to release new version soon (it's been too long) I had to do a few fixes in `OrgFormatter`. I ended up rewriting the whole links-related part. I can merge this PR when it's ready if rebase turns out to be too difficult.

aancel:
Ok for the removal of the milestone ! I'll keep on working on it as time permits (and I'll be happy with a new version too !)
Regarding to the memory issue, I think that it might come from AsyncTask that are queued up in the background. I'll have a look at the visual profiler to help debug this, thanks ! (Maybe using file.length() would help preventing too many background tasks)

I did try rebasing the PR, but yeah, it's a bit more complicated this time. I'll see how it goes, but I might end up squashing commits to make it easier on my side.

aancel:
I've given glide a try previously, but I did not seem to be working very well with ImageSpan. I can give it another try, but I'll first update the code base to the latest one.
The rebase is not going very well right now, especially in OrgFormatter.kt as there was some code refactoring (notably with FileLinkSpan), so I'm trying to update with losing anything first, before going for something new.
Right now, I can say, that due to the changes, files ending with / will be considered as files and will be linkified at first. I'll see if I can fix that later on.
Regarding ImageSpan, if we want the images to be clickable, it has to de deployed over a ClickableSpan, so over a FileLinkSpan with the changes you made.

nevenz:
> The rebase is not going very well right now, especially in OrgFormatter.kt as there was some code refactoring (notably with FileLinkSpan), so I'm trying to update with losing anything first, before going for something new.

You don't really have to rebase, you can just continue the work and I'll merge and resolve this work later. You could just work *towards* the current code in master, to make resolving conflicts easier.

Or I can modify `OrgFormatter` for what you need if it will help you. Like exposing the view for example again. I suggest moving all the code in separate classes to get a clear idea of the interface. Even better, writing an actual `interface` would be perfect. I'd like to keep `OrgFormatter` as small and generic as possible.

> Regarding ImageSpan, if we want the images to be clickable, it has to de deployed over a ClickableSpan, so over a FileLinkSpan with the changes you made.

Yes, over `FileLinkSpan`. And new `ImageLinkSpan` if we go that way. Latter could be searched for somewhere else - some place closer to the needed `TextViewWithMarkup`.

aancel:
Yes, I see what you mean, but it's just a way to make things easier on your side.
I totally agree with keeping `OrgFormatter` as simple as possible, I found the previous modifications I make are not really in this direction.

Just a quick question I have regarding the setText method of the TextView. It's just to be sure about preventing issues with the async part of image loading (I didn't find the information as of now).
When the `OrgFormatter.parse(...)` method is called, a SpannableStringBuilder is returned and set via `setText`, do you know how this will be kept in memory ? As a pure CharacterSequence or do the Span objects are kept and can be updated ?
If this is raw text, there might be an issue, if the asynchronous load of an image finishes before the call to the parse function, as text might be reset (via `setImageSpan`).

nevenz:
> When the OrgFormatter.parse(...) method is called, a SpannableStringBuilder is returned and set via setText, do you know how this will be kept in memory ? As a pure CharacterSequence or do the Span objects are kept and can be updated ?

`CharacterSequence` is an interface. It should be stored as `SpannedString`.

It can be stored as `SpannableString` (by passing `BufferType.SPANNABLE` to `setText()`) which is mutable, so that we could add more spans to it.

> If this is raw text, there might be an issue, if the asynchronous load of an image finishes before the call to the parse function, as text might be reset (via setImageSpan).

I think we could simply do:

```
spannable = OrgFormatter.parse(str)
            
textView.setParsedText(spannable)
            
imageLoader.loadImages(textView)
```

`setParsedText` would keep original text and `setText` with `BufferType.SPANNABLE`

`loadImages` images would look for `FileLinkSpan` spans and add `ImageSpan`s over them if possible/enabled/etc.

Does that sound OK? Do you need anything else apart from `TextViewWithMarkup` with `FileLinkSpan`s (that have `path`) for the `imageLoader`?

aancel:
Ok, thanks a lot for the input and explanations !
Was it a voluntary choice not to use `BufferType.SPANNABLE` with setText at first ? Or maybe is it linked to a performance issue or just the evolving use case ?

> > 
> 
> I think we could simply do:
> 
> ```
> spannable = OrgFormatter.parse(str)
>             
> textView.setParsedText(spannable)
>             
> imageLoader.loadImages(textView)
> ```

Great, it is fine for me. So if I understood well, this code would end up in the setRawText function of TextViewWithMarkup ? (unless you prefer to keep the opportunity to have images in reminders and titles)

> Does that sound OK? Do you need anything else apart from TextViewWithMarkup with FileLinkSpans (that have path) for the imageLoader?

Yes, that's OK. The FileLinkSpans will do fine. I'll have a look at this !

aancel:
Ok, so I've rebased on your latest commit to benefit from the placeholder code you added.
The code has been reworked quite a bit and it is much simpler now. I also switched to glide, which further reduces the code size and seems to have improved the loading performance for the images (and also allows not to re-invent the wheel).
The dependency attributes on the settings have also been added.

Hopefully, I didn't break anything, but eveything seems fine on API 16, 23 and 27.

aancel:
I am also keeping in mind the possibility to:
- allow the opening of a directory via a link (This can be done by modifying the intent)
- allow to set a base path for the file links (I'd prefer to put every file linked by orgzly, in subfolders of my orgzly repository)
What do you think about this ? If validated, I would maybe put the modifications in an other PR ?

nevenz:
> Ok, so I've rebased on your latest commit to benefit from the placeholder code you added.

Great, thanks!

> Hopefully, I didn't break anything, but eveything seems fine on API 16, 23 and 27.

When I open the notebook (then go back, then open again, two times) I see:

out

If I start scrolling and go back, they all end up being the same (larger) size (which is still not large enough I think).

nevenz:
> allow the opening of a directory via a link (This can be done by modifying the intent)
> allow to set a base path for the file links (I'd prefer to put every file linked by orgzly, in subfolders of my orgzly repository)
> What do you think about this ? If validated, I would maybe put the modifications in an other PR ?

Sure, both would be useful, new issue (if it needs more discussion) or PR would be great.

app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
```diff
@@ -11,17 +27,116 @@ object ImageLoader {
...
+    }
+
+    fun fitDrawable(view: View, drawable: Drawable) {
+        // Get the display metrics to be able to rescale the image if needed
+        val metrics = view.context.resources.displayMetrics
+
+        // Gather drawable information
+        // Scale the height by the scaledDensity to get original image size
+        val drawableHeight = drawable.intrinsicHeight.toFloat() * metrics.scaledDensity
+        val drawableWidth = drawable.intrinsicWidth.toFloat() * metrics.scaledDensity
```
nevenz: Isn't `scaledDensity` for text?

app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
```diff
@@ -11,17 +27,116 @@ object ImageLoader {
...
+                drawable.setBounds(0, 0, AppPreferences.setImageFixedWidth(context), AppPreferences.setImageFixedWidth(context))
+
+                Glide.with(context)
+                    .asDrawable()
+                    // Use a placeholder
+                    .apply(RequestOptions().placeholder(drawable))
+                    // And scaled thumbnails for faster display
+                    .thumbnail(Glide.with(context)
+                            .applyDefaultRequestOptions(RequestOptions().override(AppPreferences.setImageFixedWidth(context)))
+                            .load(contentUri))
```
nevenz: How can this improve display? It's the same image, no? I think this is meant to be for actual smaller images. It's likely we are now loading two same images twice (unless Glide is detecting it's the same uri and refusing to do so).

app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
```diff
@@ -11,17 +27,116 @@ object ImageLoader {
...
+            // Get the file
+            val file = File(Environment.getExternalStorageDirectory(), path)
+
+            if(file.exists()) {
+                // Get the Uri
+                val contentUri = FileProvider.getUriForFile(context, BuildConfig.APPLICATION_ID + ".fileprovider", file)
+
+                // Setup a placeholder
+                val drawable = ColorDrawable(Color.TRANSPARENT)
+                drawable.setBounds(0, 0, AppPreferences.setImageFixedWidth(context), AppPreferences.setImageFixedWidth(context))
```
nevenz: I still get the jump for the fixed size images. Looks like this is setting the height to the width, not calculating the ratio like for the actual image display.

app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
nevenz: Actually, why is this multiplied at all?

I just tried displaying a small icon (96x96), removing any other code below this, and it becomes a huge, low-quality image.

aancel:
> When I open the notebook (then go back, then open again, two times) I see:
> 
> out
> 
> If I start scrolling and go back, they all end up being the same (larger) size (which is still not large enough I think).

I also noted this behavior, it seems that the textview is not refreshed properly when changing the settings, and it does not seem to happen all the times.
I've seen that you can force a redraw with an invalidate(), but haven't had time to test it yet. Does it seem to be a good idea ?

> Sure, both would be useful, new issue (if it needs more discussion) or PR would be great.

Ok, I will open a PR then !

app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
aancel: That's a very good point indeed.
Apparently and according to the documentation the thumbnail is loaded in parallel while the main image is loading, so it should indeed load the image twice.
This was part of a test that I did. I noticed that without the thumbnail code I ended up with around 80 mb of memory used (via the visual profiler) on my sample test, and with the thumbnail code, I ended up around 38 mb.
Can you confirm this behavior ?

I would guess that the main image is not loaded and only the downsampled thumbnail. I'll dig deeper to figure out what happens behind the scene. There might be a way to save memory here.

app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
aancel: Yes, my bad.
There a few things wrong with this code indeed. The metrics.scalingDensity artifically scales the small images. The width of the view is equal to zero. The size of the Drawable are not correct regarding to the original image size (I found a way to get the original size with BitmapDrawable)
I'll rework this code

app/src/main/java/com/orgzly/android/ui/ImageLoader.kt
nevenz: Sounds like it could be the reason for random size images then? I think we should just drop `thumbnail()`.

nevenz:
I've fixed few things - preferences' dependencies, images were still being scaled up, placeholder was not being set at all. There are still some issues - images are resized twice, ratio is lost on second resize in some cases, storage permission requirement needs to be made more clear, etc.

aancel:
Thank you for the merge and thanks a lot for the fixes/testing/advices you gave on this MR !

Ypot:
I am trying to find out weeks ago:
What is the syntax to link a file that is in the external SDcard?

PR #399, opened 2018-10-15
Add support for opening org-protocol://org-id-goto links

benmoran:
I'm finding Orgzly really useful!  One feature I wanted was to be able to open notes from external links.  It's possible to do this in Emacs using `org-protocol` - https://orgmode.org/worg/org-contrib/org-protocol.html - and I've set up a handler on my PC to use `org-id-goto` so I can generate links into my Org files from a web page. 

(This is useful in conjunction with some other tools that can generate web pages summarizing Org files, for instance to help with GTD reviews.)

The attached patch allows the same links to work in Orgzly, so I can click through from a web page on my phone into the relevant note.

I appreciate it's a bit customized!  Both the `org-protocol` standard, and the `org-id-goto` functions are part of org, but `org-protocol://org-id-goto` is specific to my setup.  Is there anything else people might want to use these links for, or a more generic way of doing something like this?

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/OrgProtocolTest.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/OrgProtocolTest.java
@@ -1,0 +1,128 @@
+package com.orgzly.android.espresso;
+
+import android.content.Intent;
+import android.net.Uri;
+import android.support.test.espresso.ViewInteraction;
+import android.support.test.rule.ActivityTestRule;
+import android.view.View;
+
+import com.orgzly.R;
+import com.orgzly.android.AppIntent;
+import com.orgzly.android.NotePosition;
+import com.orgzly.android.OrgzlyTest;
+import com.orgzly.android.ui.MainActivity;
+import com.orgzly.android.ui.ShareActivity;
+
+import org.hamcrest.core.IsInstanceOf;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.action.ViewActions.click;
+import static android.support.test.espresso.assertion.ViewAssertions.matches;
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+import static com.orgzly.android.espresso.EspressoUtils.onSnackbar;
+import static com.orgzly.android.espresso.EspressoUtils.toLandscape;
+import static com.orgzly.android.espresso.EspressoUtils.toPortrait;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.startsWith;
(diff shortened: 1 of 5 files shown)


nevenz:
Cool!

I agree it would be good to be able to configure this somehow. Perhaps starting from defining supported set of actions in Settings:

- Open note with ID property
- Create new note
- etc.

For example regex could be mapped to each action. Opening note by ID accepts one parameter, so it can have regex such as `org-id-goto?id=(.*)`.

One issue - when you rotate the device, or get back to the app - same intent is handled every time and note is opened again. Check `MainActivity#setupDisplay` below it. It's where requested notebook or note are opened already - this should probably be moved there.

Thanks!

debanjum:
Hey, thanks for the really useful pull request. It'd be great if we can add the ability to open Orgzly Search Queries using the org-protocol. E.g `org-protocol://(i.todo t.work)` or based on @nevenz's suggestion `org-protocol://org-id-goto?tag=work+call&state=todo`

benmoran:
Thanks for the suggestions!  I'll have a think about how to make it flexible, while still being useful with regular Emacs as well as Orgzly.

nevenz:
Do we have some examples of `org-protocol` links people use? So that we can come up with a structure similar to majority of usages.

There's "sub-protocol" for example. Should we put all Orgzly-specific things (like query search) under `org-protocol://orgzly://` ?

benmoran:
I think the most standard use of org-protocol on the desktop is so people can make bookmarklets to easily capture snippets from their web browsers to Emacs. The main examples in the org website are aimed at that, like `org-protocol://org-capture`. 

That use case isn't really necessary on Android, because of sharing via Intents.

However org-protocol also gives a way for Emacs to define custom handlers for other links. But these will be rather unique to every user's own situation as they involve writing custom elisp functions. I don't think there is a majority use that makes sense.

My own application is a project that parses Org mode files into a neo4j database, so I can query them and make interactive plots. Then I want to link back from the visualizations to the original files, both in Emacs and Orgzly.  I'll publish this soon, maybe it will be clearer then!

benmoran:
(So having just an `org-protocol://orgzly` link wouldn't be great for me, I'd like to open these links on the desktop too. Even though I'll have to define a new Emacs function to do that, orgzly wouldn't be the right name.)

nevenz:
> However org-protocol also gives a way for Emacs to define custom handlers for other links. But these will be rather unique to every user's own situation as they involve writing custom elisp functions. I don't think there is a majority use that makes sense.

I think just a list of what's out there would be useful. Maybe it could be grepped from emacs-orgmode ML.

> (So having just an org-protocol://orgzly link wouldn't be great for me, I'd like to open these links on the desktop too. Even though I'll have to define a new Emacs function to do that, orgzly wouldn't be the right name.)

I agree, that's why I said "Orgzly-specific". I was thinking about suggested query searches.

> My own application is a project that parses Org mode files into a neo4j database, so I can query them and make interactive plots. Then I want to link back from the visualizations to the original files, both in Emacs and Orgzly. I'll publish this soon, maybe it will be clearer then!

Cool!

hh:
@benmoran how is your project coming?
> project that parses Org mode files into a neo4j database

Issue #156, opened 2017-08-22
Support to import/export settings


mjf:
Hello,

add support to import/export *Orgzly* application settings, please. Not only would it be great improvement but it would be especially helpful to keep the list of *note states* so that one wouldn't have to keep it somewhere apart on his/her own. Thank you.

-- 
*mjf*

nevenz:
Replying to @benmoran from #400:

> Do you think exporting the settings to a YAML file would be OK?

YAML's good.

> Should there be book level settings as well as global ones?

Yes, all Org mode's in-buffer settings would be per book. Most of the current Orgzly Settings would be global.

> Would it make sense to allow TODO keywords to vary across books as they do in Emacs?

Definitely - support for that has been requested in the past.

We could start by defining the schema. It'll need some care as later changes might be difficult.

Ypot:
Any hint about how the settings, searches, etc can be exported/imported?

Thx

nevenz:
> Any hint about how the settings, searches, etc can be exported/imported?

Searches can be exported to a file. As for settings, if device is not rooted Google's backup could be used. If rooted, you could grab `/data/data/com.orgzly/shared_prefs/com.orgzly_preferences.xml`.

billwanjohi:
Having done this another time now, I no longer think the status quo is quite as bad for people like myself who synchronize notes, as long as you installed Orgzly via Google Play and it's properly backed up / restored.

My most recent restore steps, which I have not yet found any gaps with:

1. Perform a cloud-based app restore, waiting on all orgzly to be restored with settings, search and notes
2. From hamburger menu on top left, select `Searches`. Then from pea-pod menu on top-right, `Export` those to a local file.
3. Again from pea-pod menu on top right, select `Settings`, then `App`, then `Clear database`. (**WARNING**: obviously don't do this if you doubt whether the synchronized versions of your notes are complete)
4. Import the searches you just exported, from same menu
5. Set up the sync repositories from scratch.

It's possible I lost some settings in the process, but I haven't noticed anything yet.

amberin:
A first version of this feature was added in [Orgzly Revived 1.9.0](https://github.com/orgzly-revived/orgzly-android-revived/releases/tag/v1.9.0).

PR #400, opened 2018-10-15
Allow Android M Direct Share to individual Orgzly books

benmoran:
Do we want to allow sharing directly to individual notebooks from Android M, using the "direct share" feature?

Perhaps we also need a way to let the user control which books should be share targets?

### DIFF
--- a/app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
+++ b/app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
@@ -12,6 +12,9 @@ import kotlinx.android.synthetic.main.activity_bookchooser.view.*
 import android.graphics.BitmapFactory
 import android.os.Bundle
 import com.orgzly.R
+import com.orgzly.android.query.Condition
+import com.orgzly.android.query.Query
+import com.orgzly.android.query.user.DottedQueryBuilder
 
 private const val DIRECT_SHARE = "ORGZLY_DIRECT_SHARE";
 
@@ -29,7 +32,8 @@ class ChooserShareTargetService : ChooserTargetService() {
                 continue;
             val bundle : Bundle = Bundle();
             val score : Float = 1.0F; // Can we guess a score based on recent use or content?
-            bundle.putString(AppIntent.EXTRA_FILTER, "b.\"" + book.name + "\"");
+            val query = DottedQueryBuilder().build(Query(Condition.InBook(book.name)));
+            bundle.putString(AppIntent.EXTRA_FILTER, query)
             targets.add(ChooserTarget(book.name, icon, score, componentName, bundle ));
         }
         return targets;
--- a/app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
+++ b/app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
@@ -13,6 +13,7 @@ import android.graphics.BitmapFactory
 import android.os.Bundle
 import com.orgzly.R
 
+private const val DIRECT_SHARE = "ORGZLY_DIRECT_SHARE";
 
 @RequiresApi(Build.VERSION_CODES.M)
 class ChooserShareTargetService : ChooserTargetService() {
@@ -23,6 +24,9 @@ class ChooserShareTargetService : ChooserTargetService() {
 
         val icon = Icon.createWithResource(this, R.drawable.ic_logo_for_widget);
         for (book in shelf.books) {
+            val direct = book.orgFileSettings.getLastKeywordValue(DIRECT_SHARE);
+            if (direct.isNullOrBlank())
+                continue;
             val bundle : Bundle = Bundle();
             val score : Float = 1.0F; // Can we guess a score based on recent use or content?
             bundle.putString(AppIntent.EXTRA_FILTER, "b.\"" + book.name + "\"");
--- a/app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
+++ b/app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
@@ -1,0 +1,34 @@
+package com.orgzly.android
+
+import android.app.PendingIntent
+import android.content.ComponentName
+import android.content.IntentFilter
+import android.graphics.drawable.Icon
+import android.os.Build
+import android.service.chooser.ChooserTarget
+import android.service.chooser.ChooserTargetService
+import android.support.annotation.RequiresApi
+import kotlinx.android.synthetic.main.activity_bookchooser.view.*
+import android.graphics.BitmapFactory
+import android.os.Bundle
+import com.orgzly.R
+
+
+@RequiresApi(Build.VERSION_CODES.M)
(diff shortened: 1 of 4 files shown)


nevenz:
I think we'll want to be able to configure the notebooks, as well as disable the feature completely. Is latter even possible? Would just returning an empty list be acceptable?

app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
```diff
@@ -0,0 +1,34 @@
...
+    private val shelf = Shelf(this)
+
+    override fun onGetChooserTargets(componentName: ComponentName?, intentFilter: IntentFilter?): MutableList<ChooserTarget> {
+        val targets = arrayListOf<ChooserTarget>();
+
+        val icon = Icon.createWithResource(this, R.drawable.ic_logo_for_widget);
+        for (book in shelf.books) {
+            val bundle : Bundle = Bundle();
+            val score : Float = 1.0F; // Can we guess a score based on recent use or content?
+            bundle.putString(AppIntent.EXTRA_FILTER, "b.\"" + book.name + "\"");
```
nevenz: This can be a problem if notebook contains quotes. Perhaps (not tested):

```suggestion
            val query = DottedQueryBuilder().build(Query(Condition.InBook(book.name)))
            bundle.putString(AppIntent.EXTRA_FILTER, query)
```

Although I don't see a reason to pass a query, maybe just a book name would be enough?

app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
```diff
@@ -0,0 +1,34 @@
...
+class ChooserShareTargetService : ChooserTargetService() {
+    private val shelf = Shelf(this)
+
+    override fun onGetChooserTargets(componentName: ComponentName?, intentFilter: IntentFilter?): MutableList<ChooserTarget> {
+        val targets = arrayListOf<ChooserTarget>();
+
+        val icon = Icon.createWithResource(this, R.drawable.ic_logo_for_widget);
+        for (book in shelf.books) {
+            val bundle : Bundle = Bundle();
+            val score : Float = 1.0F; // Can we guess a score based on recent use or content?
```
nevenz: We could use notes' created-at times to determine which notebook has the most new notes created in the last N days.

Also, list could be sorted by notebooks' modification time (like app does if that preference is used).

I'm not sure how much is all that worth the trouble though.

benmoran:
Thanks for the comments! I think it would be useful to let the user specify somehow which books should be share targets. I find that only the first two from the Share provider are shown in the Direct Share list, and currently they're arbitrary.  Perhaps the user could control this with some setting on the book.  And if no books have this setting, hopefully the feature can be disabled completely as you suggest.  I'll look into it.

What would be the best mechanism to persist these settings - perhaps a special comment in the preamble?

nevenz:
> What would be the best mechanism to persist these settings - perhaps a special comment in the preamble?

I don't think that's necessary, as it's Orgzly and Android-specific. I prefer avoiding any such custom properties, in-buffer settings etc. (https://github.com/orgzly/org-java/pull/10 is still useful for other things though).

I'd just put this in Settings.

benmoran:
> > What would be the best mechanism to persist these settings - perhaps a special comment in the preamble?
> 
> I don't think that's necessary, as it's Orgzly and Android-specific. I prefer avoiding any such custom properties, in-buffer settings etc. ([orgzly/org-java#10](https://github.com/orgzly/org-java/pull/10) is still useful for other things though).
> 
> I'd just put this in Settings.

Very true that it's Orgzly-specific.

The reason I thought it might be nice to sit in the preface:

 - It's a notebook specific setting and the settings are global, so I wasn't sure how I'd express it there
 - It avoids the issues in https://github.com/orgzly/orgzly-android/issues/156 - the preference gets exported and synced with the notebook itself, so if you delete your repo and resync it from Git, etc. then you don't need to go and fiddle around with all your settings each time.  

I was imagining next adding an option in the Notebook long-press menu that would add/remove this preface option directly.

But if you have a strong preference otherwise I can take it out and find another way.

(Actually, maybe the same preface approach could work for persisting keyword states, etc. as requested in #156, since `#+TODO: TODO(t) STARTED(s) WAITING(w) | DONE(d) CANCELED(c)` is a legitimate way of setting the keywords for a file in org-mode as well?  Though in Orgzly the states are common to all books, and not notebook specific.)

app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
benmoran: Ah, building the query properly looks nicer.

I only used a query because the `ShareActivity` already looks for a query in the `EXTRA_FILTER`:
```
                if (intent.hasExtra(AppIntent.EXTRA_FILTER)) {
                    Query query = new DottedQueryParser().parse(intent.getStringExtra(AppIntent.EXTRA_FILTER));
                    String bookName = QueryUtils.extractFirstBookNameFromQuery(query.getCondition());
...
```

Is there a simpler way to open a book with just a name?

nevenz:
I prefer not seeing `#+ORGZLY_ANYTHING` in my org files. It *is* probably the easiest solution. However, more file formats for notebooks will be supported in the future. And some might not have a good place for storing options like these.

As you mention, there's #156 and I think making this option survive notebook deletion should be part of it. I don't think we should worry about it for now.

All settings, global or per notebook, could be exported to a single user-friendly configuration file. Including settings such as `#+TODO` per notebook. Of course if notebook is exported to a format supporting any of the settings (like Org mode supports `#+TODO`), those will be included in the file (in-buffer settings in the case of org file).

Anyway, if notebook "preface" is not to be used, this setting needs to be stored somewhere else in Orgzly. Perhaps just a list of notebooks in Settings will do for now. Similar to `pref_key_share_notebook`. That will even survive notebook deletion (assuming notebook is back with the same name). And could be updated on notebook rename (though I wouldn't worry about that now).

benmoran:
Do you think exporting the settings to a YAML file would be OK?
Should there be book level settings as well as global ones? Would it make sense to allow TODO keywords to vary across books as they do in Emacs?

So many questions ;)

nevenz:
Replied to settings file stuff in #156.

app/src/main/java/com/orgzly/android/ChooserShareTargetService.kt
nevenz: > I only used a query because the ShareActivity already looks for a query in the EXTRA_FILTER

Ah right, `ShareActivity` is used for this too.

nevenz:
We can decide later how to configure this differently (if needed), as this change doesn't force creating of custom property. Whoever wants to use it can put it in the preface manually.

Thanks!

benmoran:
Thanks!  Should I make a documentation PR as well?  Or do you prefer it isn't officially documented until the settings issues are resolved?

nevenz:
@benmoran I think we keep it under radar for now.

PR #409, opened 2018-10-28
Add option to keep screen on

### LINKED ISSUE #162: Add option to runtime toggle (or switch off permanently) the "Screen Brightness Lock"
Hello.

I think an ability to toggle the *Android*'s "Screen Brightness Lock" *on* and *off* would be **very handy**!

Sometimes user would like to stare at her note for a long time and it's really annoying to keep touching the device again and again to keep the screen *on* (i.e. you lay your device on the desk and want to go through some longer note or notebook and write something on your PC at the same time).

If it can't be implemented as runtime toggle option accesible from (for example) menu anytime for some reason then provide at least global *Settings* option for it (as it is implemented i.e. in the *Acrobat Reader*).

oujunhao:
This PR resolves #162 

Sorry about using the wrong branch---I'm new to open-source but I'd like to participate more!
Let me know if I did the merging/commits wrong.


nevenz:
> Sorry about using the wrong branch---I'm new to open-source but I'd like to participate more!
Let me know if I did the merging/commits wrong.

Looks good to me, thanks!

Few questions/notes:

* Option is not displayed in query results, I think we want it there too
* Do we want this inside note details screen?
* It stays on when note details are entered, then left by pressing back. Depending on the answer to the above question that might be OK. Then again, it might be better to turn it off on any screen change (perhaps in `onPause()`).

app/src/main/java/com/orgzly/android/ui/fragments/BookFragment.java
```diff
@@ -379,6 +380,8 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
         if (BuildConfig.LOG_DEBUG) LogUtils.d(TAG, menu, inflater);
 
         inflater.inflate(R.menu.book_actions, menu);
+        boolean keepScreenOnEnabled = (getActivity().getWindow().getAttributes().flags & WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON) != 0;
+        menu.getItem(3).setChecked(keepScreenOnEnabled);
```
nevenz: This could break on menu changes, we should use `menu.findItem(R.id.books_options_keep_screen_on)`

app/src/main/java/com/orgzly/android/ui/fragments/BookFragment.java
```diff
@@ -415,6 +418,28 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 listener.onBookPrefaceEditRequest(mBook);
                 return true;
 
+            case R.id.books_options_keep_screen_on:
+                boolean keepScreenOnEnabled = (getActivity().getWindow().getAttributes().flags & WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON) != 0;
+                if (!keepScreenOnEnabled) {
+                    String title = "Keep screen on";
+                    String message = "Your screen will not turn off until you disable this option or leave this page";
```
nevenz: We'll want these in `strings.xml` so they can be translated. There's already `R.string.keep_screen_on` to be reused. And both `setTitle` and  `setMessage` below accept resource ids.

app/src/main/java/com/orgzly/android/ui/fragments/BookFragment.java
```diff
@@ -415,6 +418,28 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+                    AlertDialog.Builder builder = new AlertDialog.Builder(getActivity());
+                    builder.setTitle(title);
+                    builder.setMessage(message);
+                    builder.setPositiveButton(android.R.string.yes, (dialog, which) -> {
+                        getActivity().getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
+                        item.setChecked(true);
+                        dialog.dismiss();
+                    });
+                    builder.setNegativeButton(android.R.string.cancel, (dialog, id) -> dialog.dismiss());
+                    AlertDialog dialog = builder.create();
```
nevenz: Use can assign this to `dialog` that's already in scope, so it's dismissed in `NoteListFragment#onPause` (fixes window leak on rotation).

app/src/main/res/menu/book_actions.xml
```diff
@@ -12,6 +12,13 @@
         android:id="@+id/books_options_menu_book_preface"
         app:showAsAction="never"
         android:title="@string/edit_book_preface"/>
+    <group android:checkableBehavior="all">
+        <item
+            android:id="@+id/books_options_keep_screen_on"
+            app:showAsAction="never"
+            android:title="@string/keep_screen_on"/>
+    </group>
```
nevenz: I don't think there is a need for group here, you can just use `android:checkable="true"` on the item?

app/src/main/java/com/orgzly/android/ui/fragments/BookFragment.java
oujunhao: Thanks for the comment! I don't have much experience with android and I appreciate it. What `dialog` in scope do you mean? Something like what you have in `delete`?

```java
    private void delete(final Set<Long> ids) {
        dialog = new AlertDialog.Builder(getContext())
                .setTitle(R.string.delete_notes)
                .setMessage(R.string.delete_notes_and_all_subnotes)
                .setPositiveButton(R.string.delete, (dialog, which) -> listener.onNotesDeleteRequest(mBookId, ids))
                .setNegativeButton(R.string.cancel, (dialog, which) -> {})
                .create();

        dialog.show();
    }
```

That would mean moving the functionality to a method in `listener`, right?

Thanks again for all the help! I'm learning a lot.

oujunhao:
> Option is not displayed in query results, I think we want it there too
Ok, that's a good idea
> Do we want this inside note details screen?
I think so? Maybe because someone might want to have the contents section open
>It stays on when note details are entered, then left by pressing back.
I think the issue here would be people forgetting to turn it off and draining the battery... maybe that's a good idea to turn it off when leaving the screen.

oujunhao:
Ok, I think that's everything!

nevenz:
It doesn't seem to go away when I change screens? Looks like it's done on activity's `onPause`, not fragments'.

And it's not displayed in agenda, only plain search. Fragments are structured like this:
```
NoteListFragment
|-> BookFragment
|-> QueryFragment
    |-> SearchFragment
    |-> AgendaFragment
```

oujunhao:
Whoops, I think that last commit will take care of those issues.

nevenz:
I've moved it to `ActivityUtils` as there was some duplicated code and I've added new preference which needs to be enabled for this to show up.

Thanks!

PR #416, opened 2018-11-25
Make german introduction more complete and consistent

heyarne:
I translated some of the sections in the german introduction that were still in English and I proof-read it and changed it for some improved consistency. Hope this helps, thanks for the app!


heyarne:
I just saw that I would have to change the file line by line in crowdin, so I'd have to go over the complete file and check my changes again to copy and paste them. I don't have time to do this right now unfortunately. Is there maybe a possibility to re-import it into crowdin?

nevenz:
> Is there maybe a possibility to re-import it into crowdin?

Doesn't look like it, unless it's CSV or XML file.

I made some changes going through the diff. Not all ("Note can include links" section), as it's based on an older original.

Thanks again.

Issue #80, opened 2017-05-15
Org-Capture Implementation


Atrus7:
Howdy,

I'd love to have org-capture within orgzly, and I've been looking into implementing it. I've poked around and worked on it a bit, and I think it's time to check in and ask a few questions..

I plan on building it such that it works with voice commands. 

"Ok Google, capture [idea]" 
This would prompt the user to begin speaking and would fill in the idea template. 

The template would be a set of default templates or user-edited ones. Currently I plan on using a subset of the template syntax/features that are used in the emacs version of org-capture. Certain template options are either not applicable to implement or would be too complex for my contribution's scope.

Questions:
@nevenz  is this something that you'd be interested in pulling into your repo if it was done?
Have you already pursued this in a long-forgotten branch? What do you think? Any problems you foresee?

nevenz:
There have been requests for implementing templates, so yeah, actually implementing templates sounds great. 😄

BTW, you can already quickly create a note:
- Using voice: Saying "create note" and using Orgzly as an app (probably can be set as default)
- Enabling ongoing notification, so you can just swipe down notification drawer and click

But you can only change the notebook currently and note will be created as a last top-level note in it. So choosing a template should probably be squeezed somewhere in there.

I don't know what's possible to do using voice. Can we actually define a phrase and map it to activity/action/extras? If not, in the worst case, Orgzly can recognize it in the title. So instead of saving a note using "capture orgzly feedback save current location as a property" as a title, Orgzly can create "Save current location as a property" in "orgzly.org/Feedback/". If we defined "orgzly feedback" as a template. Something along those lines?

Using voice or not, a new dropdown can be added to `ShareActivity`. Currently there is only one for changing the notebook. Adding one for location was planned (to allow browsing the tree, like when refiling in Org mode). Another one can be for templates which, when item is selected, would set both notebook and location, as well as any other property (state, priority, title, tags, planning times, properties, content). Though location should probably include the notebook ("orgzly.com/Feedback").

Anyway, just random thoughts.

I'd probably start with data structure and location - how and where to store defined templates.

tkamat:
I already have this setup to use voice using IFTTT. 

For the "if this" section in the builder, choose the google assistant assistant trigger "say a simple phrase with a text ingredient." Customize the trigger phrase(s) to fit your liking, and enter the response you want.

[...]

PR #466, opened 2019-02-24
Basic org-capture implementation

karlicoss:
Related issue: https://github.com/orgzly/orgzly-android/issues/80 

This is a very minimalistic implementation, basically, just a shorthand to creating a note in a specific notebook. But it'd be a necessary step towards implementing proper capture anyway, so I think it makes sense to do it now considering that actually supporting templates would be a bit more complicated.

In my vision, in future when we do support templates, you'd just have a UI prompting you for notebook and also asking you to choose from one of the predefined capture templates. Then the template is passed to the share intent and handled by the fragment that creates notes (e.g. filling up default todo state/tags/etc). If you don't recieve a template in the intent, you just create a note in the target notebook. So the behaviour would be backwards compatible.

There is a fair bit of copy-pasting in `CaptureChooserActivity.onBookClicked` from `BookChooserActivity.onBookClicked`.. but not too much. Let me know if you think it should be generified.

Let me know if there is anything you think needs improvement here. If this all reasonable, I'll implement some tests too, just want to make sure this approach looks fine to you.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -105,8 +105,8 @@
         </activity>
 
         <activity
-            android:name="com.orgzly.android.ui.CaptureChooserActivity"
-            android:label="@string/capture_shortcut_label"
+            android:name="com.orgzly.android.ui.TemplateChooserActivity"
+            android:label="@string/new_note_shortcut_label"
             android:icon="@mipmap/cic_shortcut_new_note">
 
             <intent-filter>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -105,6 +105,17 @@
         </activity>
 
         <activity
+            android:name="com.orgzly.android.ui.CaptureChooserActivity"
+            android:label="@string/capture_shortcut_label"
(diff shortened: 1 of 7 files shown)


nevenz:
Looks like a good start. I'd just rename "capture" to "template" everywhere, I think it's more clear? Widget can be "New from template" or something?

karlicoss:
Can't say it would be much more clear, but on the other hand 'capture' is not inherently obvious too. So I don't mind renaming! For widget, perhaps even 'New note' is clear enough?

nevenz:
> For widget, perhaps even 'New note' is clear enough?

Sure, sounds good.

karlicoss:
ok, done!

karlicoss:
Hey @nevenz wonder if we it looks good and you can merge that? :)

nevenz:
Thanks!

We might want to rename widget names, as we now have:

- New note
- Orgzly notebook
- Orgzly notes

They were prefixed by "Orgzly" as older Android versions are not grouping widgets by app. That's probably not a good reason any more to have redundant app name, so we could probably have something like:

- New note
- Open Notebook
- Saved search

PR #480, opened 2019-03-06
Notify on reminders by default

licaon-kter:
...I feel this is to be expected.


nevenz:
Notification sounds are currently played one after another when multiple notifications are fired at once (for daily tasks for example). I think this needs to be changed before notification are enabled by default, as it's very annoying.

licaon-kter:
Thanks

PR #492, opened 2019-03-14
Ukrainian translation

MichaelKazarian:
First of all, thank you for really great work! I want to add my 5 cents and translate orgzly into Ukrainian.


nevenz:
Thanks!

Any chance you can upload it to https://crowdin.com/project/orgzly/uk# ? I can do it if you don't already have an account (I've never tried that, but it seems possible).

MichaelKazarian:
Today I created an account in https://crowdin.com/ uploaded and updated translations. Here new translations.

PR #500, opened 2019-03-17
add option to remove common indent from note body

derrotebaron:
For example:
```
* foo
** bar
   example content
     deeper indent
```
will display
```
example content
  deeper indent
```
instead of
```
   example content
     deeper indent
```
as text of a note.

This does not apply to the EditText, because it might confuse the user when
typing spaces at the beginning of a line, as they wouldn't show up if they are
part of a common indent.

The reasoning behind this feature is, that indents in Emacs Org-Mode are by
default as deep as needed to match the text of the innermost subheading 
that contains the body to be indented.

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/SettingsChangeTest.java
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/SettingsChangeTest.java
@@ -12,10 +12,13 @@ import org.junit.Rule;
 import org.junit.Test;
 
 import static androidx.test.espresso.Espresso.onData;
+import static androidx.test.espresso.Espresso.onView;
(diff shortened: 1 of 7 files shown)


nevenz:
This is when `org-adapt-indentation` is used?

I think this should be an option under "Settings / Sync / Org file format" and only done when parsing or writing to Org files. I don't think content should be changed within the app.

It could be detected automatically too (currently done for planning times and properties [when parsing](https://github.com/orgzly/org-java/blob/cef586fb390277fb78d46827000466bef81c493a/src/main/java/com/orgzly/org/parser/OrgSaxyParser.java#L184)), though that might not always work.

Issue #320, opened 2018-04-27
remove "TODO" from items in widget TODO view.


naught101:
It would be good if there was a way to remove the "TODO" prefix on todos in the widget, because they block most of the text, and make the actual list unreadable, especially when the widget is relatively narrow. I'll add a screenshot.

naught101:
![phimpme1524804958707](https://user-images.githubusercontent.com/167164/39345480-fbbf81fe-4a2b-11e8-9019-0c68af687913.png)

NiceFeather:
Acturally all of widget can be resize in the most of home lanucher, long press&resizing it will be OK.

naught101:
Yes, I know, but I would prefer for the widget to not take up the wholes screen. For example, my current layout looks like this: 

![screenshot_20180427-145432](https://user-images.githubusercontent.com/167164/39384968-e28fee18-4ab1-11e8-9294-5fdf3c5b863c.png)

I would like to keep the TODO list on my main home screen, but if it is any bigger than this, then it takes up too much space and my home screen becomes less useful as a whole.

NiceFeather:
Yes, @nevenz maybe set an option for this state to show/hide is more flexible like 'Inherited tags in search results' of tag, this is just a proposal.

BTW, the option of 'Inherited tags in search results' above seems has a bug, all note tags always show up.

![screenshot_20180428-114811](https://user-images.githubusercontent.com/28109556/39393521-b946719a-4ab5-11e8-984f-33de62ccfd03.jpg)

nevenz:
Would removing checkmark be enough? Note would be selectable instead and action menu would open. I'm not a fan of allowing state to be removed as it's pretty important part of the note. But if we must, I guess it could be allowed for the widget only (as a part of #142).

> BTW, the option of 'Inherited tags in search results' above seems has a bug, all note tags always show up.

Note's yes, but its ancestors' shouldn't.

With option off:

![off](https://user-images.githubusercontent.com/10671476/39430621-974d795a-4c8e-11e8-9825-6c7d50a64499.png)

With option on:

![on](https://user-images.githubusercontent.com/10671476/39430680-c189716a-4c8e-11e8-8a5a-a06ecf87df3c.png)

naught101:
Re: state removal: I would only suggest that for a state-based search, I guess. Everything having the same state makes the state information redundant.

Another option might be to replace "TODO" with a small logo, or just a small "T".

[...]

Issue #207, opened 2017-11-03
Widget status toggle


skandranii:
"Widget status toggle" for the notes .. is kinda risky cause it accidentally can switch a note from todo to Done without noticing if you just touched by mistake.

nevenz:
That button could be removed (it takes a lot of space anyway) and I think the list behaviour should be consistent to app's (selecting notes & quick-menu).

PR #514, opened 2019-04-03
Widget: show check mark for NOTEs as well

lolilolicon:
More convenience.
There is no reason to block the ability to mark notes "done" from the widget.


nevenz:
This is kinda going in opposite direction from #207 and #320. Perhaps user should be able to choose for which types of states he wants the checkmark displayed. Some might want it for done notes too, to "uncheck" them.

Thanks!

PR #539, opened 2019-05-02
Update strings.xml in russian translation

dvmi:
Disambiguation between 'refile' and 'move' in translations.


nevenz:
Translations are done on https://crowdin.com/project/orgzly, files are not directly modified. I can update it there if you can't or don't want to use it.

Thanks!

dvmi:
Thank you.
I've updated translation at crowdin.

PR #491, opened 2019-03-13
Working Git repositories

amberin:
OK, so, first of all... I'm not really a programmer, much less a Java programmer. I just really wanted this functionality myself. I used this as a kind of exercise to learn something about Java and app development.

I've probably made some horrible decisions here, but this code seems to work well for me (on a Samsung S7), and I'm hoping that someone more experienced will be able to suggest improvements if this is not acceptable. I'd also be willing to re-work things if I can get pointers in the right direction.

I believe the commit messages explain more or less all my changes.

PR #173, opened 2017-09-12
Add Git Synchronization

IvanMalison:
Fixes #24

Issue #24, opened 2017-02-27
git synchroniztion


IvanMalison:
The possibility of new synchroniztion methods was mentioned way back in 2015 https://plus.google.com/+Orgzly/posts/S2nBbxVibsB

Did anything become of this? Are there plans to add git synchronization support to orgzly?

angrybacon:
Google Drive synchronization is wanted too.

nevenz:
You can open a separate issue for Google Drive sync, thanks!

dogear42:
@IvanMalison 

When you say git synchronization, were you just talking about committing to a local repo, or also potentially pulling/pushing as part of the sync from that repo to a remote?

I'm just wondering what the full scope of this feature would be.

I like the idea of having the "git client" handle the conflicts during a replay as part of pull-before-push, perhaps adding a new commit with the conflict left in it verbatim, so that I can resolve it later.  Like:

```
 <<<<<<< HEAD:mergetest
 This is my third line
 =======
 This is a fourth line I am adding
 >>>>>>> 4e2b407f501b68f8588aa645acafffa0224b9b78:mergetest
```

I think this would be a huge improvement from the limited options of force-load/force-save that Orgzly currently provides.

jappeace:
Shouldn't this be done by a separate program? Or for example make a library from the core features of: https://github.com/maks/MGit and then use those in this project to provide nice integration.

mkaito:
I use Resilio for phone-to-computer sync. Orgzly-to-sdcard is the kind of auto sync I want. Git might be a good backend for conflict resolution, but staying in sync automatically will outright prevent many conflicts from ever happening.

El 3 de mayo de 2017 7:15:32 GMT+01:00, Jappie Klooster <notifications@github.com> escribió:
>Shouldn't this be done by a separate program? Or for example make a
>library from the core features of: https://github.com/maks/MGit and
>then use those in this project to provide nice integration.

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

angrybacon:
@mkaito That would prevent the user to edit a file that is not up to date ie. no offline mode.

mkaito:
I think the constant need to manually synchronize are a bigger concern than the occasional lack of connection.

I'm not saying there won't ever be conflicts, just that most happen when you forget to hit the sync button.

El 3 de mayo de 2017 16:04:03 GMT+01:00, Mathieu Marques <notifications@github.com> escribió:
[...]

PR #543, opened 2019-05-13
Git sync base changes

lfxgroove:
Hi again, this is related to #24.

@nevenz, @IvanMalison could one of you please take a look at this draft to just get things working with regards to syncing a git repository? I'm working off the list mentioned in https://github.com/orgzly/orgzly-android/pull/173#issuecomment-336303192 but wanted to check that the base I'll be working on isn't busted first.

A few things I'm not sure about:
* We handle an initially empty git repo by checking for a RevCommit that is null or a getObjectId() that is null, this leads to a few strange checks, see e.g. https://github.com/orgzly/orgzly-android/commit/b3736b224a789580fbc1cd5c28ed6eafa7478b2a#diff-fed1654ff699f68e117dce23ae1625acR157 where null means that we should push.
* DataRepository is now sent as an argument to both GitRepo and RepoPreferences, maybe there's a better way to solve this?
* I'm not sure how sync back should be handled, currently SyncService#handleTwoWaySync checks if loadFile is null and in that case doesn't do much.
* I've made GitFileSynchronize#updateAndCommitFile() public since it seemed to fit what I wanted to do when there are no commits in the repo, thoughts?

Regarding authentication, I'd like to add support for generating a SSH-key as an alternative to user/pass authentication, should I do that in a later PR or can I cram that into this one?

Keeping the list of things to fix here, from #173:

**Issues or missing features to be resolved before enabling this in the app**

* [x]  Warning when syncing from Android Studio

```
Warning:WARNING: Dependency org.apache.httpcomponents:httpclient:4.1.3 is ignored for fdroidDebug as it may be conflicting with the internal version provided by Android.
```

* [x]  Building fails when doing `./gradlew build` (`proguard-rules.pro` needs updating probably)

* [x]  Asking for Storage permission _before_ entering Git repo creation, leaving user in repos list, see 596ff7f9e604241093011b77df0c9d95d8b73465

* [ ]  Prepopulate directory (Orgzly/repos/git in storage root or similar)
Semi-done, currently a directory is created when none is selected and is auto-selected. But I'm not sure that's what we want to do if the user decides to setup another repo as we'll try to re-use that same folder in that case which won't work out.

* [x]  On tapping checkmark (using `git@bitbucket.org:nevenz/foo.git`)

```
11-09 08:05:24.383  7914  7914 W System.err: java.io.IOException: Directory /storage/emulated/0/Download/git is not a git repository.
11-09 08:05:24.383  7914  7914 W System.err:    at com.orgzly.android.repos.GitRepo.ensureRepositoryExists(GitRepo.java:109)
11-09 08:05:24.384  7914  7914 W System.err:    at com.orgzly.android.repos.GitRepo.ensureRepositoryExists(GitRepo.java:73)
11-09 08:05:24.384  7914  7914 W System.err:    at com.orgzly.android.ui.fragments.GitRepoFragment$RepoCloneTask.doInBackground(GitRepoFragment.java:343)
11-09 08:05:24.384  7914  7914 W System.err:    at com.orgzly.android.ui.fragments.GitRepoFragment$RepoCloneTask.doInBackground(GitRepoFragment.java:312)
```
As far as I can tell this has been solved.

* [ ]  Had to delete selected (empty, created from browser) directory (can't use just browser, we have to type?)

* [ ]  Edit repo, change to non-existing URL, no check (dialog) this time, repo is accepted

* [ ]  Trying to sync gives "No repos configured" (with non-existing URL created above). Later discovered that the file from repository was downloaded at one point after all

* [x]  "Unsupported repository type" trying to edit repo (non-existing URL)

```
11-09 08:15:50.019  7914  7914 W System.err: java.io.IOException: The file /storage/emulated/0/Documents/nevenz/foo2.git does not exist
11-09 08:15:50.019  7914  7914 W System.err:    at com.orgzly.android.repos.GitRepo.ensureRepositoryExists(GitRepo.java:105)
11-09 08:15:50.019  7914  7914 W System.err:    at com.orgzly.android.repos.GitRepo.ensureRepositoryExists(GitRepo.java:73)
11-09 08:15:50.019  7914  7914 W System.err:    at com.orgzly.android.repos.GitRepo.build(GitRepo.java:57)
11-09 08:15:50.019  7914  7914 W System.err:    at com.orgzly.android.repos.GitRepo.buildFromUri(GitRepo.java:53)
11-09 08:15:50.019  7914  7914 W System.err:    at com.orgzly.android.repos.RepoFactory.getFromUri(RepoFactory.java:49)
11-09 08:15:50.019  7914  7914 W System.err:    at com.orgzly.android.repos.RepoFactory.getFromUri(RepoFactory.java:23)
11-09 08:15:50.019  7914  7914 W System.err:    at com.orgzly.android.ui.ReposActivity.onRepoEditRequest(ReposActivity.java:188)
11-09 08:15:50.019  7914  7914 W System.err:    at com.orgzly.android.ui.fragments.ReposFragment.onListItemClick(ReposFragment.java:130)
```
Should be fixed, can't reproduce

* [ ]  Using `https://nevenz@bitbucket.org/nevenz/foo.git`

```
10-12 23:20:20.008  6287  6416 E AndroidRuntime: Caused by: java.lang.ClassCastException: org.eclipse.jgit.transport.TransportHttp cannot be cast to org.eclipse.jgit.transport.SshTransport
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at com.orgzly.android.git.GitSSHKeyTransportSetter$2.configure(GitSSHKeyTransportSetter.java:40)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at org.eclipse.jgit.api.TransportCommand.configure(TransportCommand.java:138)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at org.eclipse.jgit.api.FetchCommand.call(FetchCommand.java:128)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at org.eclipse.jgit.api.CloneCommand.fetch(CloneCommand.java:193)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at org.eclipse.jgit.api.CloneCommand.call(CloneCommand.java:133)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at com.orgzly.android.repos.GitRepo.ensureRepositoryExists(GitRepo.java:91)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at com.orgzly.android.repos.GitRepo.ensureRepositoryExists(GitRepo.java:73)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at com.orgzly.android.ui.fragments.GitRepoFragment$RepoCloneTask.doInBackground(GitRepoFragment.java:343)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at com.orgzly.android.ui.fragments.GitRepoFragment$RepoCloneTask.doInBackground(GitRepoFragment.java:312)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at android.os.AsyncTask$2.call(AsyncTask.java:305)
10-12 23:20:20.008  6287  6416 E AndroidRuntime:        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
```

* [x]  Clicking checkmark immediately (all fields empty): Snackbar with `javaj.io.IOException: Failed to clone repository , null`

* [x]  Entering "asd" as URL and pressing checkmark

```
java.io.IOException: Failed to clone repository asd, org.eclipse.jgit.errors.NoRemoteRepositoryException: asd: not found.
```
this is fixed, however, maybe we should replace the progress modal with a progress bar of some kind?

* [ ]  Error checking for all other fields

* [ ]  Allow login with user/pass

* [ ] Update JGit version to a newer one and solve the entrySet() problem

**Found during testing/changes from @amberin**
* [ ] Changing the storage location for a repo makes the repo unusable
* [ ] mtime correction
* [ ] Change how conflict resolution is handled, push the conflicting merge branch to remote to allow fixing it somewhere else and then allow "force-pulling" a notebook. This probably doesn't fit in this PR but I wanted to remember it.

**Minor stuff**

* [x]  Big button for Git needs to be added in repos list (between Dropbox and Directory)

* [x]  Change order in `+` drop-down (Dropbox, Git, Directory)

* [x]  Can't scroll down (not using ScrollView?)

* [ ]  Git branch's input is half a size (top half is cut off)

* [ ]  Snackbar displayed below opened keyboard

* [x]  Action button (`imeOptions`) - `actionNext` for all but last, which should be `actionDone`

* [x]  Remove "Git" from all hints

* [ ]  Git author (e.g. My Name) -> Name (e.g. John Doe)

* [x]  Git email (e.g. [me@email.com](mailto:me@email.com)) -> Email (e.g. [john@doe.com](mailto:john@doe.com))

* [x]  Git branch (e.g. master) -> Branch

* [x]  Prepopulate branch with "master"

* [ ]  Fields with browse button don't have hint displayed when value is set (`TextInputLayout` missing?)

* [ ]  Git remote address -> Remote URL

* [ ]  Extract string resources displayed to user (errors etc.)

* [ ]  Move git package to sync/ or repos/ (for now)

Thanks!


IvanMalison:
@lfxgroove awesome! Thanks for doing this! I'll take a look soon!

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
```diff
@@ -237,8 +241,14 @@ public RevCommit currentHead() throws IOException {
     public RevCommit getCommit(String identifier) throws IOException {
         Log.i("test", git.getRepository().getWorkTree().toString());
```
IvanMalison: Did I leave these here? Seems like they should probably be removed.

app/src/main/java/com/orgzly/android/repos/GitRepo.java
```diff
@@ -146,6 +164,18 @@ RevCommit getCommitFromRevisionString(String revisionString) throws IOException
         return walk().parseCommit(ObjectId.fromString(revisionString));
     }
 
+    // TODO: Should this be in the synchronizer instead?
+    RevCommit getCommitFromFileName(String fileName) throws IOException {
```
IvanMalison: Why do we ever need to get the commit for a filename?

app/src/main/java/com/orgzly/android/repos/GitRepo.java
```diff
@@ -129,11 +138,20 @@ public VersionedRook storeBook(File file, String fileName) throws IOException {
 //        VersionedRook current = CurrentRooksClient.get(
 //                // TODO: get rid of "/" prefix needed here
 //                App.getAppContext(), getUri().toString(), "/" + fileName);
-        VersionedRook current = null;
+        //VersionedRook current = null;
+        //RevCommit commit = getCommitFromRevisionString(current.getRevision());
+
+        RevCommit commit = getCommitFromFileName(fileName);
```
IvanMalison: it seems like the only place that this is used is in getFileRevision. In that case, shouldn't we just be able to use whatever HEAD is as the commit?

app/src/main/java/com/orgzly/android/sync/SyncService.java
```diff
@@ -536,18 +536,21 @@ private static void handleTwoWaySync(DataRepository dataRepository, TwoWaySyncRe
...
-                    bookName.getFormat(),
-                    result.getLoadFile(),
-                    newRook);
-            // TODO: db.book().updateIsModified(bookView.book.id, false)
-            // Instead of:
-            // dataRepository.updateBookMtime(loadedBook.getBook().getId(), 0);
+            // We only need to write it if syncback is needed
+            if (result.getLoadFile() != null) {
+                newRook = result.getNewRook();
+                String fileName = BookName.getFileName(App.getAppContext(), newRook.getUri());
```
IvanMalison: I'm not really familiar with what is going on here. Seems like things might have changed quite a bit since I originally wrote git sync.

app/src/main/java/com/orgzly/android/repos/GitRepo.java
lfxgroove: I've removed the getCommitFromFileName function and now keep HEAD as a reference before doing any changes so that tryPushIfUpdated still works as expected. Does that sound ok?

app/src/main/java/com/orgzly/android/repos/GitRepo.java
lfxgroove: Solved with your second comment, this function is removed now

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
lfxgroove: I've removed them now :)

lfxgroove:
I'll take the review as I'm not totally off the mark, so I'll keep going down the checklist from #173. Along with this i'll look into adding support for generating a ssh-keypair as an alternative to user/pass. Let me know if there are any other issues I should keep a lookout for or if I should change my approach!

app/src/main/java/com/orgzly/android/sync/SyncService.java
lfxgroove: Is @nevenz someone who knows their way around this code, or should I poke someone else? I'll try to get a better understanding myself as well.

IvanMalison:
Hey, so I checked out the repo again to try to understand some of whats going on here, and I noticed this:

https://github.com/orgzly/orgzly-android/blob/ad0330c102ab08d636b7fe1441dfdc5f9f4a2e9f/app/src/main/java/com/orgzly/android/repos/GitRepo.java#L128

https://github.com/orgzly/orgzly-android/blob/ad0330c102ab08d636b7fe1441dfdc5f9f4a2e9f/app/src/main/java/com/orgzly/android/repos/GitRepo.java#L152

https://github.com/orgzly/orgzly-android/blob/ad0330c102ab08d636b7fe1441dfdc5f9f4a2e9f/app/src/main/java/com/orgzly/android/repos/GitRepo.java#L156

It's not entirely clear to me what @nevenz  means here by 

```
removed current_versioned_rooks table, just get the list from remote
```

This is probably the most important thing to fix from the perspective of synchronization. IMO its super important that orgzly store some sort of hash of the contents of the file at last synchronization. Without this its not really possible to make sure that updates are applied in a non destructive way.

Anyway, this needs to be resolved in order for us to proceed. @lfxgroove How clearly do you feel like you understand the design of GitRepo/GitSynchronizer and why things were done the way they are? One of the things I really wanted out of the git synchornization was to be able to rely on gits own merge/conflict resolution abilies, but in order for this to work we have to be very careful about how we apply updates to files. Just want to make sure you understand what is "allowed" and what isn't in order for this to work properly.

IvanMalison:
@lfxgroove maybe we could chat somewhere not in this thread about how to proceed?

IvanMalison:
@nevenz In case its not clear, what I would like to know is:

How should we go about finding/storing meta data about our last synchronization of a file with a remote. I was storing hashes that allowed us to determine whether what was currently stored was in git history.

lfxgroove:
> 
> 
> @lfxgroove maybe we could chat somewhere not in this thread about how to proceed?

Sure, do you have an IRC channel, Matrix perhaps? I'm not really sure what people use, so just give me a program and some channel or something to connect to and I should be able to log on at some point.

> 
> Anyway, this needs to be resolved in order for us to proceed. @lfxgroove How clearly do you feel like you understand the design of GitRepo/GitSynchronizer and why things were done the way they are? One of the things I really wanted out of the git synchornization was to be able to rely on gits own merge/conflict resolution abilies, but in order for this to work we have to be very careful about how we apply updates to files. Just want to make sure you understand what is "allowed" and what isn't in order for this to work properly.

I'm not that clear on what's allowed and not, I can try to read up on the code a bit more to get a better understanding of it. But it sure would be nice if you could clarify what you mean, but like you said, perhaps somewhere else :)

nevenz:
@IvanMalison, there's `BookSync` for every book that's synced, which connects book with `VersionedRook`. `VersionedRook` could be extended if its `rookRevision` is not enough.

I'm not sure if that's helpful. I'll have to get on top of this issue again if needed.

IvanMalison:
@lfxgroove Looking at it again, IIRC, because the GitRepo class supplies two way sync, it basically does not need to implement storeBook and retrieveBook at all.

The current syncBook method handles everything correctly as far as I can tell, so we should just delete those methods @nevenz correct me if I am wrong on that point.

nevenz:
> The current syncBook method handles everything correctly as far as I can tell, so we should just delete those methods

I think they are still used (at least `storeBook` should be) when there is no remote notebook. It's in that check in `SyncService#syncNamesake`.

amberin:
> > The current syncBook method handles everything correctly as far as I can tell, so we should just delete those methods
> 
> I think they are still used (at least `storeBook` should be) when there is no remote notebook. It's in that check in `SyncService#syncNamesake`.

That's right. It was also discussed in my PR: https://github.com/orgzly/orgzly-android/pull/491#discussion_r266249250

amberin:
@lfxgroove,
> DataRepository is now sent as an argument to both GitRepo and RepoPreferences, maybe there's a better way to solve this?
Not sure if you are aware, but I passed in the repo ID instead. Not saying that was better -- just want to make sure you are aware of the possibility. (I'd also love to hear any reasons for not doing that.)

lfxgroove:
@amberin Wasn't really aware of this no, I didn't notice that you added a function RepoPreferences :) To me this sounds more reasonable than passing the whole DataRepository in. I'll change to that, any objections @IvanMalison ?

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
```diff
@@ -154,7 +155,7 @@ private boolean doMerge(RevCommit mergeTarget) throws IOException, GitAPIExcepti
     }
 
     public void tryPushIfUpdated(RevCommit commit) throws IOException {
-        if (!commit.equals(currentHead())) {
+        if (commit == null || !commit.equals(currentHead())) {
```
IvanMalison: Not sure i understand this change, can you explain why we need it?

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
```diff
@@ -103,7 +103,8 @@ public boolean updateAndCommitFileFromRevisionAndMerge(
         if (updateAndCommitFileFromRevision(sourceFile, repositoryPath, fileRevision)) return true;
 
         String originalBranch = git.getRepository().getFullBranch();
-        String mergeBranch = String.format("merge%s%s", repositoryPath, fileRevision.getName());
+        // TODO: Why do we want to give the branch a magic name? Can we do the equivalent of git merge origin/master while we are on master?
```
IvanMalison: The main reason we use a special branch name for merge commits is because we need to give the user some way to resolve conflicts when they do arise.

I think that there is an error in the code as it exists, though because doCleanup is always set to true. Really, what should happen is that if we see that the merge failed, but we did not get a GitAPIException, we should set doCleanup to be false. This will make it so that orgzly's current state will be properly synced back to SOME branch (so orgzly's updates will have made it into git), but it won't be the original branch that orgzly was syncing with. This means that the user is expected to manually merge the new branch that was created back in to the original branch, and change head to point to the correct thing eventually, using some external git manipulation.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
```diff
@@ -103,7 +103,8 @@ public boolean updateAndCommitFileFromRevisionAndMerge(
...
+        String mergeBranch = String.format("merge%s%s", repositoryPath.replace(" ", ""), fileRevision.getName());
         try {
             git.branchDelete().setBranchNames(mergeBranch).call();
```
IvanMalison: One thing that I failed to consider before when I was writing this is that this could be somewhat dangerous in the case where we have already had a merge issue with this file. In that case, we could be deleting the branch that we are currently on!

This is bad and needs to be changed. we should add a third component, which is maybe just the current date (or perhaps a uuid which would be the safest thing).

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
```diff
@@ -196,6 +197,9 @@ public void setBranchAndGetLatest() throws IOException {
             // FIXME: maybe:
             // checkoutSelected();
             RevCommit current = currentHead();
+            if (current == null) {
```
IvanMalison: Why is this needed?

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
```diff
@@ -208,7 +212,7 @@ public void setBranchAndGetLatest() throws IOException {
         }
     }
 
-    private RevCommit updateAndCommitFile(
+    public RevCommit updateAndCommitFile(
```
IvanMalison: As mentioned in another comment, I think this should remain private.

app/src/main/java/com/orgzly/android/git/GitPreferencesFromRepoPrefs.java
```diff
@@ -1,16 +1,20 @@
...
 import com.orgzly.android.prefs.RepoPreferences;
 
 public class GitPreferencesFromRepoPrefs implements GitPreferences {
     private RepoPreferences repoPreferences;
+    private DataRepository repo;
 
-    public GitPreferencesFromRepoPrefs(RepoPreferences prefs) {
+    public GitPreferencesFromRepoPrefs(RepoPreferences prefs, DataRepository repoIn) {
         repoPreferences = prefs;
+        repo = repoIn;
```
IvanMalison: Hmmmm. this feels a little untidy to me.

This object only exists to expose information about the users git preferences...

IvanMalison: Maybe just pass the id iteself in since that is all that is needed?

app/src/main/java/com/orgzly/android/repos/GitRepo.java
```diff
@@ -40,14 +46,26 @@
 public class GitRepo implements SyncRepo, TwoWaySyncRepo {
     public final static String SCHEME = "git";
 
+    /**
+     * Used as cause when we try to clone into a non-empty directory
+     */
+    public static class DirNotEmpty extends Exception {
```
IvanMalison: This is Java, no character skimping.

app/src/main/java/com/orgzly/android/repos/GitRepo.java
```diff
@@ -40,14 +46,26 @@
...
+    }
+
     public static GitTransportSetter getTransportSetter(GitPreferences preferences) {
         return new GitSSHKeyTransportSetter(Uri.parse(preferences.sshKeyPathString()).getPath());
     }
 
-    public static GitRepo buildFromUri(Context context, Uri uri)
+    public static GitRepo buildFromUri(Context context, Uri uri, DataRepository repo)
             throws IOException, URISyntaxException {
+        Log.d("mylog", "Got uri " + uri.toString());
```
IvanMalison: Remove this log.

app/src/main/java/com/orgzly/android/sync/SyncService.java
```diff
@@ -536,18 +536,21 @@ private static void handleTwoWaySync(DataRepository dataRepository, TwoWaySyncRe
...
-            Log.i("Git", String.format("Loading from file %s", result.getLoadFile().toString()));
-            BookView loadedBook = dataRepository.loadBookFromFile(
-                    bookName.getName(),
-                    bookName.getFormat(),
-                    result.getLoadFile(),
-                    newRook);
-            // TODO: db.book().updateIsModified(bookView.book.id, false)
-            // Instead of:
-            // dataRepository.updateBookMtime(loadedBook.getBook().getId(), 0);
+            // We only need to write it if syncback is needed
```
IvanMalison: I forget, are we storing the file revision, or the full git sha as the remote book ID. If its the latter, we should make sure we are doing the right thing in that we are still updating our view of the remote appropriate.

app/src/main/java/com/orgzly/android/ui/repo/GitRepoActivity.kt
```diff
@@ -90,6 +123,25 @@ class GitRepoActivity : CommonActivity(), GitPreferences {
                 activity_repo_git_url.setText(repo.url)
                 setFromPreferences()
             }
+        } else {
+            createDefaultRepoFolder();
+        }
+    }
+
+    // TODO: Since we can create multiple syncs, this folder might be re-used, do we want to create
```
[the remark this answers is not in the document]
amberin: I have been wanting to get rid of folder selection in the UI and instead simply creating a folder (preferably in the app's private storage) named after the repo ID. But that would probably require a temporary path for the initial cloning, since we don't know the repo ID until the new repo has been saved to the database.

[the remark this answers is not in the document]
IvanMalison: @amberin, not sure I want to force the use of app private storage, since it will be important to surface access to the git repository to the user so they can resolve merge conflicts etc.

gradle.properties
```diff
@@ -65,7 +65,7 @@ evernote_android_job_version = 1.2.6
 
 joda_time_version = 2.10.1
 
-jgit_version = 4.11.5.201810191925-r
+jgit_version = 4.4.1.201607150455-r
```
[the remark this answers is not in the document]
lfxgroove: 79775af84d tries to explain it a bit. The problem with the currently used version is that it tries to create a hardlink at clone time, which Android doesn't allow. Newer versions solve this by feature checking if hardlink is possible, this old version also doesn't attempt hardlinking.

When i tried using a newer version I got a runtime error for HashMap.entrySet() iirc. I can add a todo in the list up top to investigate this properly and see if I can get it working.

lfxgroove:
> Bunch of thoughts here. I'd work on it myself but getting android studio up and running is a pain, and I hate life outside of emacs.

Thanks for the comments! I'm not sure how you want to do this, for me it's okay with some comments as i push commits. But I guess if you want to do less work totally I could ping you once I consider everything done, some things are placeholders for now to get stuff working while I figure other stuff out.

app/src/main/java/com/orgzly/android/ui/repo/GitRepoActivity.kt
```diff
@@ -351,6 +425,56 @@ class GitRepoActivity : CommonActivity(), GitPreferences {
...
+        }
+
+        override fun doInBackground(vararg params: Void): IOException? {
+            // TODO: Ask the user for a passphrase and size of the key we generate
+            return try {
+                val jsch = JSch()
+                val keyPair = KeyPair.genKeyPair(jsch, KeyPair.RSA, 4096)
+
+                // TODO: Do we want constants for the file names?
+                val privKeyFile = fragment.applicationContext.openFileOutput(".ssh_key", Context.MODE_PRIVATE)
```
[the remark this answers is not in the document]
lfxgroove: I'm not sure how we want this feature to work. Do we want the user to be able to associate separate keys for each sync they setup or do we just offer one key that is the Orgzly key so to speak? This was just to make testing for me easy, I didn't put much thought into it.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
lfxgroove: This is for the case when we don't have a HEAD, i.e. we've cloned a empty repository, then the commit we get in will be null (from what I can remember). However, I guess it might make more sense to check if `currentHead() == null` as that should imply the same thing?

app/src/main/java/com/orgzly/android/ui/repo/GitRepoActivity.kt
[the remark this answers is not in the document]
IvanMalison: Not really sure. I think we should cut as much work from v1 as possible. Just one key would work for me. If someone needs that feature they can come ask for it...

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
IvanMalison: My question would be: Do we need to hit this code path at all in the case where we are cloning a new repository? I haven't thought about it... that's a real question. In an ideal world I'd like to avoid using null whenever possible.

IvanMalison:
I'll keep looking at it periodically. Seems like this change is getting reasonably big.

app/src/main/java/com/orgzly/android/ui/repo/GitRepoActivity.kt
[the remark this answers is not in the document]
lfxgroove: Imo it doesn't make sense to put the generated keys in the home folder, so i'll keep them in the storage for Orgzly for now, but they're renamed to `id_rsa{.pub}`. The reason I don't want to use the home directory is since we might use a key the user didn't expect if we're not clear about it, and for now it seems easier to keep track of our own key.

app/src/main/java/com/orgzly/android/git/GitPreferencesFromRepoPrefs.java
lfxgroove: If I understood the comment for remoteUri() it needs the repository to be able to fetch the URI? Or can I get that information from the RepoPreferences somehow?

app/src/main/java/com/orgzly/android/repos/GitRepo.java
lfxgroove: Done

app/src/main/java/com/orgzly/android/repos/GitRepo.java
lfxgroove: Done

app/src/main/java/com/orgzly/android/git/GitPreferencesFromRepoPrefs.java
IvanMalison: ahh i see.

app/src/main/java/com/orgzly/android/ui/repo/GitRepoActivity.kt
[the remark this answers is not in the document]
lfxgroove: Resolving for now, tell me if you want this solved some other way

app/src/main/java/com/orgzly/android/git/GitPreferencesFromRepoPrefs.java
lfxgroove: This dependency has been removed by passing in the URL instead as amberin did.

app/src/main/java/com/orgzly/android/repos/GitRepo.java
```diff
@@ -125,16 +181,16 @@ public boolean requiresConnection() {
...
-//        VersionedRook current = CurrentRooksClient.get(
-//                // TODO: get rid of "/" prefix needed here
-//                App.getAppContext(), getUri().toString(), "/" + fileName);
-        VersionedRook current = null;
-
-        RevCommit commit = getCommitFromRevisionString(current.getRevision());
-        synchronizer.updateAndCommitFileFromRevision(
-                file, fileName, synchronizer.getFileRevision(fileName, commit));
-        synchronizer.tryPushIfUpdated(commit);
+        RevCommit prevCommit = null;
```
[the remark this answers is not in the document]
lfxgroove: From what I can tell in SyncService.java this function is called when the status is one of the following:
```
ONLY_BOOK_WITHOUT_LINK_AND_ONE_REPO:
BOOK_WITH_LINK_LOCAL_MODIFIED:
ONLY_BOOK_WITH_LINK:
```
which doesn't mean that we always add a _new_ file, but rather that we don't have any conflicts with the remote repo if one exists? I just want to make sure that we're on the same page. But what you're saying should still be valid as far as I can tell so I'll give implementing it a go.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
lfxgroove: With `GitFileSynchronizer#addAndCommitNewFile` this null check isn't neede anymore and the parameter here is now `@NonNull`

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
lfxgroove: Done

app/src/main/java/com/orgzly/android/repos/GitRepo.java
lfxgroove: I think this code is removed now, marking resolved.

app/src/main/java/com/orgzly/android/sync/SyncService.java
lfxgroove: VersionedRook's created in GitRepo use the SHA of the commit that created the file, there's already a call to `updateBookLinkAndSync` so it should be fine?

app/src/main/java/com/orgzly/android/repos/GitRepo.java
```diff
@@ -180,16 +180,10 @@ public boolean requiresConnection() {
...
-        if (synchronizer.isEmptyRepo()) {
-            // TODO: Is this ok to do?
-            synchronizer.updateAndCommitFile(file, fileName);
-        } else {
-            prevCommit = synchronizer.currentHead();
-            synchronizer.updateAndCommitFileFromRevision(
-                    file, fileName, synchronizer.getFileRevision(fileName, prevCommit));
-        }
-        synchronizer.tryPushIfUpdated(prevCommit);
+        // Since GitRepo implements TwoWaySync this method is only called if fileName doesn't exist
```
IvanMalison: Cool, this is basically what I had imagined.

Nit: Hard to tell from just looking at this review, but I don't think that ensureRepoIsClean (there ia also a typo in that method name that should get fixed) is called anywhere along this code path. I think it should probably get added to addAndCommitNewFile

app/src/main/java/com/orgzly/android/ui/repo/GitRepoActivity.kt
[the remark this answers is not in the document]
amberin: Ah, I see. I have always imagined that conflict resolution would take place on another system, but that of course requires the branch to get pushed to the remote, and I assume that never happens?

app/src/main/java/com/orgzly/android/repos/GitRepo.java
lfxgroove: Fixed typo and added call to the function

app/src/main/java/com/orgzly/android/ui/repo/GitRepoActivity.kt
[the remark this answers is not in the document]
IvanMalison: @amberin, yes, conflict resolution DOES likely happen on another system, but even in that case, we need to be able to manipulate the raw git repository, because when there IS a conflict, we will need to put the changes from orgzly somewhere back out there (i.e. on a branch that is not matched up with master). Then those conflicts need to be resolved and synced back in to the main master branch.

I suppose that we could have orgzly simply push its view of master to another branch, while leaving its view of the world as the local master, but I still think there is value to exposing the underlying git repository to the user just in case things get kind of hairy.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
lfxgroove: I added the current date and time as the third component.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
lfxgroove: The doCleanup part is fixed as far as i can tell. I can't install a git client on my simulator to verify this though.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
lfxgroove: Removed, was used earlier in how storeBook() was implemented.

app/src/main/java/com/orgzly/android/ui/repo/GitRepoActivity.kt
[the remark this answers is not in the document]
lfxgroove: I think the missing part here is to be able to force-pull master for a specifik notebook? Or is it supposed to be solved some other way? My thinking is that the resolution i come to on my computer may still be in conflict with the version i have on my phone, thus leading to another merge-conflict.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
IvanMalison: yeah, that's probably fine, though i still think a hash or uuid would be ideal.

IvanMalison:
@nevenz @lfxgroove I think we should maybe consider merging this before we have done ALL of the things that need to get done for git. The review and comment history are already getting reasonably unwieldy and I think we have most of the things done here that get things working again. Most of what remains is tweaks to the UI and that can probably be done in a separate review. Thanks for picking this up @lfxgroove

nevenz:
I'm fine with merging if you both and @amberin agree, I didn't have time to follow the work too closely. I'll take a quick look before merging and later go through everything more closely before enabling it (`git.enable` in `app.properties`).

Should we close #491?

amberin:
> Should we close #491?

Yes, please do.

I also agree that this PR should be merged. I'm very happy that @lfxgroove has gotten so much done.

app/src/main/java/com/orgzly/android/ui/repos/ReposActivity.kt
```diff
@@ -175,6 +178,19 @@ class ReposActivity : CommonActivity(), AdapterView.OnItemClickListener {
         }
     }
 
+    override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<String>, grantResults: IntArray) {
```
IvanMalison: @nevenz Needs to review this.

app/src/main/java/com/orgzly/android/ui/views/AdaptableHeightViewPager.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A [ViewPager] which supports changing height depending on the current child being viewed.
+ * The adapter needs to be inherit from [androidx.fragment.app.FragmentPagerAdapter] for this
+ * to work.
+ *
+ * @see ArrayPagerAdapter
+ * @see com.orgzly.android.ui.repo.GitRepoActivity for example use
+ */
+class AdaptableHeightViewPager : ViewPager {
```
IvanMalison: I didn't review this.

app/src/main/java/com/orgzly/android/ui/views/ArrayPagerAdapter.kt
```diff
@@ -0,0 +1,35 @@
...
+import android.util.Log
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentPagerAdapter
+
+/**
+ * Simple {@link androidx.viewpager.widget.PagerAdapter} implementation which can contain an array
+ * of fragments.
+ */
+class ArrayPagerAdapter : FragmentPagerAdapter {
```
IvanMalison: Didn't review this.

gradle.properties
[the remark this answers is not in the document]
IvanMalison: okay, I'm still super skeptical of this. 2016 is a long time ago.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
```diff
@@ -126,14 +151,16 @@ public boolean updateAndCommitFileFromRevisionAndMerge(
...
+                if (!result.getMergeStatus().isSuccessful()) {
                     throw new IOException("Unexpected failure to merge branch");
+                }
             }
         } catch (GitAPIException e) {
             doCleanup = true;
             e.printStackTrace();
             throw new IOException(
                     String.format("Failed to handle merge correctly: %s", e.getMessage()));
+            // TODO: want to catch CheckoutConflictException as well, that means that the actual merge produced conflicts
```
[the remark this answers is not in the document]
lfxgroove: I think CheckoutConflictException is the exception one actually wanted to catch here, while a GitAPIException implies an unknown error during merge. I'm not sure though.

lfxgroove:
@IvanMalison @nevenz The one big thing still left from the list of things to do is to support both SSH and HTTPS authentication, should we do that in another PR? Tell me if that's the case and I'll clean-up changes related to that functionality in my branch and make it ready for merge

app/src/main/java/com/orgzly/android/ui/repo/git/GitRepoActivity.kt
```diff
@@ -140,27 +209,51 @@ class GitRepoActivity : CommonActivity(), GitPreferences {
...
+                e.cause is NotSupportedException -> R.string.git_clone_error_uri_not_supported
+                else -> R.string.git_clone_error_unknown
+            }
+            showSnackbar(errorId)
             e.printStackTrace()
-            showSnackbar(e.toString())
         }
     }
 
+    // TODO: Finish this function
```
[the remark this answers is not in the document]
lfxgroove: The function has been removed for now, will probably use it in PR for both auth methods.

app/src/main/java/com/orgzly/android/git/GitFileSynchronizer.java
```diff
@@ -235,9 +277,11 @@ public RevCommit currentHead() throws IOException {
...
 
     public RevCommit getCommit(String identifier) throws IOException {
-        Log.i("test", git.getRepository().getWorkTree().toString());
-        Log.i("test", identifier);
-        Ref target = git.getRepository().getRef(identifier);
+        if (isEmptyRepo()) {
+            return null;
+        }
+        Ref target = git.getRepository().findRef(identifier);
+        Log.d("mylog", "Targetid : " + target.getObjectId());
```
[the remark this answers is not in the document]
lfxgroove: Done

app/src/main/java/com/orgzly/android/ui/repo/git/GitRepoActivity.kt
```diff
@@ -351,9 +447,61 @@ class GitRepoActivity : CommonActivity(), GitPreferences {
         }
     }
 
+    internal inner class SSHKeypairGenerationTask(var fragment: GitRepoActivity) : AsyncTask<Void, Void, IOException>(), ProgressMonitor {
```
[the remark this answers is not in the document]
lfxgroove: Removed

app/src/main/java/com/orgzly/android/ui/repo/git/GitRepoActivity.kt
```diff
@@ -140,27 +209,51 @@ class GitRepoActivity : CommonActivity(), GitPreferences {
...
         }
     }
 
     private fun repoCheckComplete(e: IOException?) {
         if (e == null) {
             save()
-
         } else {
+            val errorId = when {
+                // TODO: Handle the private/public key being wrong or the username/password
```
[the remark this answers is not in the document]
lfxgroove: Kind of done, left a note about showing error for username/password once we have https support.

app/src/main/java/com/orgzly/android/ui/views/AdaptableHeightViewPager.kt
lfxgroove: Removed, will include in new PR with https support

app/src/main/java/com/orgzly/android/ui/views/ArrayPagerAdapter.kt
lfxgroove: Removed, will include in new PR with https support

nevenz:
Yeah, the simpler this is the best, so we can merge it and continue from there.

I gave it a try -- I've added a test repo with two files, synced, repo was cloned, I made a local modification to one file, synced and:

- for a modified notebook I get "Both local and remote notebook have been modified"
- for the notebook I didn't touch I get "Loaded from..."

In repo I see two new "Orgzly update" commits, one for each file.

I expected "Saved to" for a modified notebook.

Change for the other was white-space only, so that might be expected (even though local notebook hasn't been modified, generated org is different from the one in remote notebook).

Also, hitting sync again, I get "Loaded from" for the modified file (and again after every sync).

lfxgroove:
Hi again, I've been on vacation. I'll take a look at your comments and try to solve them, just wanted to let you know.

IvanMalison:
@lfxgroove @nevenz Just a heads up, I'm going to try working on this locally a bit. Hopefully none of my work conflicts with yours @lfxgroove

lfxgroove:
@IvanMalison, @nevenz:
I think I'm a little lost on how to go about fixing what nevenz described. I tried a little bit of what you described and ended up with the following:
have two files in the repo, a and b

change `a` on computer, push
sync on orgzly
change `b` on orgzly
sync on orgzly
=> `b` receives CONFLICT_BOTH_BOOK_AND_ROOK_MODIFIED (Both local and remote notebook have been modified)

another sync resolves this. As far as I can tell I can reproduce this over and over by doing what i described, and it seems to be the same thing as what you got @nevenz. A wild guess would be that the VRook somehow incorrectly decides that since we've done a local change and the last revision didn't include any changes for us (`b`), only for `a`, we're out of sync?

Any thoughts or ideas?

IvanMalison:
IIRC CONFLICT_BOTH_BOOK_AND_ROOK_MODIFIED is decided before we even get to the repo implementation. It was part of the reason that we had to use TwoWaySync for git repos anyway.

It seems weird that we can even get in to that state at all because we should be hitting this code path no matter what with two way sync:

https://github.com/orgzly/orgzly-android/blob/211bf478caaa70ac8f06463b7380a98cacfecde4/app/src/main/java/com/orgzly/android/sync/SyncService.java#L451

IvanMalison:
This is really an issue with the fact that orgzly assumes that state is necessarily a problem. In git's case, it is actually often possible to handle this situation. This is where we set the state:
https://github.com/orgzly/orgzly-android/blob/211bf478caaa70ac8f06463b7380a98cacfecde4/app/src/main/java/com/orgzly/android/sync/BookNamesake

amberin:
> another sync resolves this. As far as I can tell I can reproduce this over and over by doing what i described, and it seems to be the same thing as what you got @nevenz. A wild guess would be that the VRook somehow incorrectly decides that since we've done a local change and the last revision didn't include any changes for us (`b`), only for `a`, we're out of sync?

I had the same issues, and I found GitRepo#currentVersionedRook() to be the culprit. My solution looked like this https://github.com/amberin/orgzly-android/commit/cabe012af04139249e251e599307985c90b89afe#diff-6887dc2ebf7c9e65784782c2717ffcaeR166. (It used a new method in GitRepo: https://github.com/amberin/orgzly-android/commit/c82c8af08a1693eb1e25adcdfa2f2c448cb3b903#diff-fed1654ff699f68e117dce23ae1625acR278.)

IvanMalison:
@amberin yeah I think something was changed significatnly (code was maybe commented out by @nevenz that used to work) for finding/setting the rook

IvanMalison:
@amberin I think we had quite a bit of discussion about that peice of code IIRC. I don't think we should just blindly get the latest commit of the file to solve this. We want to get the actual rook that was last synced.

amberin:
> @amberin I think we had quite a bit of discussion about that peice of code IIRC. I don't think we should just blindly get the latest commit of the file to solve this. We want to get the actual rook that was last synced.

Yeah, I'm sure my code had lots of gaping holes. For what it's worth though, I've been using it daily for 3 months, and the sync has yet to misbehave. :)

IvanMalison:
@lfxgroove Do you feel like you have an understanding of what the issue is here, or do you want me to poke around a bit?

gradle.properties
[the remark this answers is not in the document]
IvanMalison: Okay so basically https://www.eclipse.org/forums/index.php/t/1096721/ explains this. The thing is that it seems like really we need to use something even older than we are using here.

IvanMalison:
I'm getting

`
java.io.IOException: Failed to clone repository git@bitbucket.com:ivanmalison/org.git, git@bitbucket.com:ivanmalison/org.git: invalid privatekey:
`

every time.

IvanMalison:
@nevenz Can you clarify what you mean in these fixmes:

https://github.com/orgzly/orgzly-android/blob/211bf478caaa70ac8f06463b7380a98cacfecde4/app/src/main/java/com/orgzly/android/repos/GitRepo.java#L203

nevenz:
> Can you clarify what you mean in these fixmes:

The link is invalid, is it https://github.com/orgzly/orgzly-android/pull/543#issuecomment-492413857 ?

lfxgroove:
> @amberin I think we had quite a bit of discussion about that peice of code IIRC. I don't think we should just blindly get the latest commit of the file to solve this. We want to get the actual rook that was last synced.

I've looked a little bit more at this. The current code kind of does the same thing as well, right? So we'd somehow like to get the latest commit that we know we have pushed when we call currentVersionedRook()? If that's the case that should be available in the database, right? And we can just double-check that what we get from the db matches what we have locally. Or am I going about this the wrong way?

IvanMalison:
@lfxgroove No, actually I was wrong. In the case of currentVersionedRook, we SHOULD giving back the state of the git repository on disk.

The reason we see the bug nevenz describes is that the commit hash will update when ANY of the relevant git files update, but orgzly syncs files independently, 1 by 1. This is why when we return currentVersionedRook, we actually need to find the latest commit that actually modified the file in question. Basically, we should just use amberin's solution.

I think that I briefly considered storing the sha of the blob of each file, rather than the commit hash, to avoid this issue, the reason I decided that was a bad idea, is that the blob sha only refers to the contents of the file, and not necessarily the point in history, so if e.g. you reverted some change in between, you could be allowed to get inconsistencies. Obviously this is a crazy corner case, but using the commit sha feels more correct.

IvanMalison:
@lfxgroove as far as fixing the issue, just use @amberin 's solution

lfxgroove:
@amberin your fix seems to work well for me as well, thanks :)
@nevez, @IvanMalison Is there something else that needs to be cleaned up? As far as I can tell the sync shows reasonable states for my two test notebooks.

nevenz:
Have force-save and force-load been tried? I get:

<details>
<summary>Force-save exception</summary>
<p>

```
07-03 15:26:03.432  2064  2927 W System.err: java.io.IOException: Can't add new file file.org that already exists.
07-03 15:26:03.432  2064  2927 W System.err:    at com.orgzly.android.git.GitFileSynchronizer.addAndCommitNewFile(GitFileSynchronizer.java:249)
07-03 15:26:03.432  2064  2927 W System.err:    at com.orgzly.android.repos.GitRepo.storeBook(GitRepo.java:186)
07-03 15:26:03.432  2064  2927 W System.err:    at com.orgzly.android.data.DataRepository.saveBookToRepo(DataRepository.kt:151)
07-03 15:26:03.433  2064  2927 W System.err:    at com.orgzly.android.data.DataRepository.forceSaveBook(DataRepository.kt:110)
07-03 15:26:03.433  2064  2927 W System.err:    at com.orgzly.android.usecase.BookForceSave.run(BookForceSave.kt:10)
07-03 15:26:03.433  2064  2927 W System.err:    at com.orgzly.android.usecase.UseCaseRunner.run(UseCaseRunner.kt:35)
07-03 15:26:03.434  2064  2927 W System.err:    at com.orgzly.android.ui.main.SyncFragment.lambda$run$2$SyncFragment(SyncFragment.java:351)
07-03 15:26:03.434  2064  2927 W System.err:    at com.orgzly.android.ui.main.-$$Lambda$SyncFragment$5nMBnY8w5-oNCPqyBt6TPFCsoIk.run(Unknown Source:4)
07-03 15:26:03.435  2064  2927 W System.err:    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
07-03 15:26:03.435  2064  2927 W System.err:    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
07-03 15:26:03.435  2064  2927 W System.err:    at java.lang.Thread.run(Thread.java:764)
07-03 15:26:03.436  2064  2927 D com.orgzly.android.data.DataRepository: 1590#AsyncTask #15: forceSaveBook: Updating status for 3
07-03 15:26:03.444  2064  2927 D com.orgzly.android.data.DataRepository: 1590#AsyncTask #15: setBookLastActionAndSyncStatus: Updating status for 3: 1
07-03 15:26:03.444  2064  2927 W System.err: java.io.IOException: Force-saving failed: Can't add new file file.org that already exists.
07-03 15:26:03.444  2064  2927 W System.err:    at com.orgzly.android.data.DataRepository.forceSaveBook(DataRepository.kt:127)
07-03 15:26:03.444  2064  2927 W System.err:    at com.orgzly.android.usecase.BookForceSave.run(BookForceSave.kt:10)
07-03 15:26:03.444  2064  2927 W System.err:    at com.orgzly.android.usecase.UseCaseRunner.run(UseCaseRunner.kt:35)
07-03 15:26:03.444  2064  2927 W System.err:    at com.orgzly.android.ui.main.SyncFragment.lambda$run$2$SyncFragment(SyncFragment.java:351)
07-03 15:26:03.444  2064  2927 W System.err:    at com.orgzly.android.ui.main.-$$Lambda$SyncFragment$5nMBnY8w5-oNCPqyBt6TPFCsoIk.run(Unknown Source:4)
07-03 15:26:03.445  2064  2927 W System.err:    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
07-03 15:26:03.445  2064  2927 W System.err:    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
07-03 15:26:03.445  2064  2927 W System.err:    at java.lang.Thread.run(Thread.java:764)
```
</p>
</details>

<details>
<summary>Force-load exception</summary>
<p>

```
07-03 15:28:11.654  2064  2955 W System.err: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.eclipse.jgit.revwalk.RevCommit.equals(org.eclipse.jgit.lib.AnyObjectId)' on a null object reference
07-03 15:28:11.655  2064  2955 W System.err:    at com.orgzly.android.git.GitFileSynchronizer.tryPushIfUpdated(GitFileSynchronizer.java:185)
07-03 15:28:11.655  2064  2955 W System.err:    at com.orgzly.android.repos.GitRepo.retrieveBook(GitRepo.java:218)
07-03 15:28:11.656  2064  2955 W System.err:    at com.orgzly.android.data.DataRepository.loadBookFromRepo(DataRepository.kt:1428)
07-03 15:28:11.656  2064  2955 W System.err:    at com.orgzly.android.data.DataRepository.forceLoadBook(DataRepository.kt:79)
07-03 15:28:11.657  2064  2955 W System.err:    at com.orgzly.android.usecase.BookForceLoad.run(BookForceLoad.kt:10)
07-03 15:28:11.658  2064  2955 W System.err:    at com.orgzly.android.usecase.UseCaseRunner.run(UseCaseRunner.kt:35)
07-03 15:28:11.660  2064  2955 W System.err:    at com.orgzly.android.ui.main.SyncFragment.lambda$run$2$SyncFragment(SyncFragment.java:351)
07-03 15:28:11.661  2064  2955 W System.err:    at com.orgzly.android.ui.main.-$$Lambda$SyncFragment$5nMBnY8w5-oNCPqyBt6TPFCsoIk.run(Unknown Source:4)
07-03 15:28:11.663  2064  2955 W System.err:    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
07-03 15:28:11.664  2064  2955 W System.err:    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
07-03 15:28:11.665  2064  2955 W System.err:    at java.lang.Thread.run(Thread.java:764)
07-03 15:28:11.684  2064  2955 D com.orgzly.android.data.DataRepository: 1591#AsyncTask #16: setBookLastActionAndSyncStatus: Updating status for 3: 1
07-03 15:28:11.685  2064  2955 W System.err: java.io.IOException: Force-loading failed: Attempt to invoke virtual method 'boolean org.eclipse.jgit.revwalk.RevCommit.equals(org.eclipse.jgit.lib.AnyObjectId)' on a null object reference
07-03 15:28:11.685  2064  2955 W System.err:    at com.orgzly.android.data.DataRepository.forceLoadBook(DataRepository.kt:92)
07-03 15:28:11.685  2064  2955 W System.err:    at com.orgzly.android.usecase.BookForceLoad.run(BookForceLoad.kt:10)
07-03 15:28:11.685  2064  2955 W System.err:    at com.orgzly.android.usecase.UseCaseRunner.run(UseCaseRunner.kt:35)
07-03 15:28:11.686  2064  2955 W System.err:    at com.orgzly.android.ui.main.SyncFragment.lambda$run$2$SyncFragment(SyncFragment.java:351)
07-03 15:28:11.686  2064  2955 W System.err:    at com.orgzly.android.ui.main.-$$Lambda$SyncFragment$5nMBnY8w5-oNCPqyBt6TPFCsoIk.run(Unknown Source:4)
07-03 15:28:11.686  2064  2955 W System.err:    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
07-03 15:28:11.686  2064  2955 W System.err:    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
07-03 15:28:11.686  2064  2955 W System.err:    at java.lang.Thread.run(Thread.java:764)
```
</p>
</details>

Also, I don't get that failed-sync-icon displayed on sync conflict.

BTW, I'm fine with merging this any time if it would make things easier. Although discussion would then need to be moved someplace else, a new PR or existing issue perhaps.

IvanMalison:
> BTW, I'm fine with merging this any time if it would make things easier. Although discussion would then need to be moved someplace else, a new PR or existing issue perhaps.

Yes, lets merge, take stock of all remaining issues and try to finish with one last PR.


> Have force-save and force-load been tried? I get:

No but based on those stack traces I have a good understanding of why they don't work and what we need to do to fix them.

caleb-allen:
@lfxgroove how is this branch looking? Any tasks I could take?

IvanMalison:
@caleb-allen branch is merged. feel free to do anything in the todo list.

PR #570, opened 2019-07-01
changed one translation

mox601:
As an italian speaker, I think it sounds better


nevenz:
Translations are done on [Crowdin](https://crowdin.com/translate/orgzly/7/en-it#q=repos_preference_summary), but I can update it if you don't (want to) have an account there.

Thank you!

mox601:
I just suggested the translation in crowdin!

Issue #38, opened 2017-03-12
WebDAV sync support


nevenz:
This will also support ownCloud (#28) and Nextcloud (#33) sync.

btimofeev:
This will also support [Yandex Disk](https://disk.yandex.ru) sync.

rafaelleru:
Working on this also.

gour:
@rafaelleru hiya, just switched back to orgmode (from taskwarrior) and looking what is the best or most reliable syncing client for Android...for my calendars/contacts I'm using Baikal, so wonder if that should be supported as well in Orgzly by this 'WebDAV sync support' ? Any (rough) ETA?

rafaelleru:
Yeah, walways that you have a webDAV link it should works fine.

gour:
@rafaelleru that's great! how far have you arrived with it?

rafaelleru:
I'm currently finishing my university exams but in one week I will have so much time, so I think that I will have it finish for august.

jmgrosen:
@rafaelleru What's your status on this? Could you push your existing code somewhere?

rafaelleru:
Hy guis I was busy this summer, on monday I will have my work pc again and upload the full code, I was restarting all my work because sardine causes problems in new android versions, I will kept you info about progress :smile:

michaelcadilhac:
Hi there @rafaelleru !  Has this been put upstream?  Do you need betatesters?  In my experience, FolderSync+Orgzly just doesn't work; Orgzly always reports sync conflicts (though I'm inclined to think that it's FolderSync's fault).  For the moment, I manually "Force save" every now and then, which is quite problematic.  Thanks so much for your work!  Cheers!

rafaelleru:
Hy, recently my pc crashes I will back to work the next week as soon I have another one, sorry for the inconvenience. I will push my work soon.

MBuchaus:
I will also volunteer to beta test this.. I'm anxiously anticipating this feature.

bighi:
So... any news on WebDAV support? Any way I can help?

uGeek:
any news on WebDAV support?

many free software applications include synchronization with nextcloud

Simpletask, joplin... 

Include webdav includes being able to synchronize with Nextcloud, Owncloud, Seafile (opensource) anda Yandex, Box.com, Stack, Zimbra... 

Many clouds with only one added. a lot of synchronization power

matt812:
[...]

PR #580, opened 2019-07-18
#38 WebDAV sync support

bemyak:


nevenz:
Great, thanks!

Works for me after some testing. I'm going to release v1.7.9 soon, then merge this for v1.8.

bemyak:
Quick summary of the latest changes:

- Resolve merge conflicts
- Added "Test" button to UI. The result message is robust, but pretty ugly, sorry :) Maybe someone who is better in UI can fix it's representation.
- Add translations of new strings to Russian

If you have any comments on implementation please let me know.

nevenz:
> Add translations of new strings to Russian

These are done on https://crowdin.com/project/orgzly, so they will be overwritten.

Worked fine for me during the little testing I did. We'll probably want to support `http(s)` too and I can fix the messages. I like the test button, we should do that for git too to avoid doing anything on new repo create.

Anyway, I can merge it if it's OK with you? Thanks!

bemyak:
In fact, we do support `http(s)`, you can enter the url in that format if you like. If will be automatically translated into `webdav(s)`.

Yes, feel free to merge this! There are some new conflicts though, I'be able to fix them only in ~12 hours.

nevenz:
> In fact, we do support http(s), you can enter the url in that format if you like. If will be automatically translated into webdav(s).

You can't go back to the repo though due to "Unsupported repository type" (due to scheme checks in `RepoFactory`).

> Yes, feel free to merge this! There are some new conflicts though, I'be able to fix them only in ~12 hours.

It's fine, I can fix those.

bemyak:
> `Unsupported repository type`
Fixed! It was a really stupid bug :)

nevenz:
> Fixed!

Thanks. We'll probably want to leave the URL as it is in the future, but again, let's wait for a different way to determine a repository type.

snuffop:
I think I may capture many's emotion with this comment.   "yay  friggin' awesome!!!!"

PR #582, opened 2019-07-29
Enable orgzly to receive shared image

lvic:
Thanks for this app and for making it open source. I love orgzly. To quickly annotate/schedule photos or screenshots I needed orgzly to allow me to share image files with it. So I added the code for this functionality. 
I built and tested it with success on my device. 

Thanks again.

P.S.: Line 6, 10 and 11 in AndroidManifest.xml are not needed leftovers inserted while getting the app to run on my device beside the 'real' orgzly app. Sorry for leaving them in the commit. They can be removed. Let me now if I shall add a new commit to remove them. I ended up using applicationIdSuffix ".debug" inside the debug buildType in the build.gradle file instead.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -3,8 +3,11 @@
 <manifest
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
+    xmlns:dist="http://schemas.android.com/apk/distribution"
     package="com.orgzly"
     android:installLocation="auto">
+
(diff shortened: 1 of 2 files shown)


nevenz:
Cool, thanks!

Perhaps a default title should be added, for quicker sharing?

BTW, indentation seems to be off.

nevenz:
I've changed/fixed few things. Thanks again.

BTW, getting path from content URI looks like a hit and miss. Probably the better option would be to actually copy the image and link to that (by default, with an option to just link if path can be figured out).

Issue #577, opened 2019-07-10
Pull-to-refresh to sync


dareenzo:
- [x] I have searched for existing issues that may be the same as or related to mine.

I'd like to suggest using PTR to sync all notebooks

PR #618, opened 2019-10-13
Pull to sync

bemyak:
Closes https://github.com/orgzly/orgzly-android/issues/577


nevenz:
Thanks!

It would be better if the icon keeps spinning while the sync is in progress, but it might be tricky to do that. I think this is fine for now.

We might want to do this in notebook view too?

BTW, can you move setting of the listener to the end of `onViewCreated()`. I've recently cleaned `onCreateView()` from everything but inflating (364f34f423fb7738a1636143ee48aeb10b70c506).

bemyak:
Yes, I've tried to make icon spin as long the the sync goes, but then realized that I need to scan activity list constantly to monitor when Sync Activity will disappear. So I decided to keep thing simple: we still have icon in notification area so it should be obvious to the user when sync ends.

I'll try to insert it into notebook view now, thank you for the suggestion.

Sure, I'll move it!

bemyak:
Done!

It was a little bit tricky, for some reason I had to wrap `RecyclerView` in `LinearLayout` to make `SwipeRefreshLayout` work. Some details on this maybe found here: https://stackoverflow.com/a/38392908

Also, I can't compile project locally without bumping `minSdkVersion` to 19. I can commit and push my changes in `build.gradle` as well if you want.

nevenz:
One more thing - this doesn't work when there are no notebooks or notes in a notebook? We'll have to support that too for consistency.

> Also, I can't compile project locally without bumping minSdkVersion to 19.

How are you compiling and what's the message?

bemyak:
I'm using `build` task in Android studio and getting this error during `app:lint` stage:
```
orgzly-android/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java:782: Error: Field requires API level 19 (current min is 16): android.os.Environment#DIRECTORY_DOCUMENTS [NewApi]
          File path = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOCUMENTS);
                                                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

bemyak:
Meanwhile got it working for empty lists. Thank you for the good catch!

Btw, I had a repo configured in orgzly v.1.7.9 and now on `master` I'm getting "No repos configured" error when I'm trying to sync. The error is gone if the repo is deleted and created again. Maybe this is something you would want to check before release.

nevenz:
I can't trigger sync in notebook in one emulator due to large value in `setDistanceToTriggerSync`. It will probably be the issue on a lot devices, especially in landscape mode. I don't think this can be a constant. It would have to depend on height.

What was the reason for changing the default? Were you triggering sync by accident when scrolling?

Perhaps we should remove it for now and see how it goes.

> orgzly-android/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java:782: Error: Field requires API level 19 (current min is 16): android.os.Environment#DIRECTORY_DOCUMENTS [NewApi]

Ah yes, this needs fixing, it's not worth increasing the min API version. I'll check it out.

> Btw, I had a repo configured in orgzly v.1.7.9 and now on master I'm getting "No repos configured" error when I'm trying to sync

Repo type has been added, but I didn't notice any issues with migrations. I'll see to test that more, thanks.

bemyak:
Yes, I thought that users scroll more on that view, so it might be a bit annoying to get that "rotating thing" coming out all the time. Maybe I set it to a bit high value, it took me all-screen-height gesture to trigger sync.

For now I removed that setting completely. You are right, we should just see how it goes and probably get some feedback.

nevenz:
> I had a repo configured in orgzly v.1.7.9 and now on master I'm getting "No repos configured" error when I'm trying to sync

@bemyak, which repo type was this?

bemyak:
@nevenz , WebDAV one

nevenz:
> WebDAV one

Ah OK cool - location of username/password preferences changed, so that's expected.

bemyak:
Great that you found it! Thanks

PR #623, opened 2019-10-22
Convert SyncService into Kotlin

bemyak:
+Increase heap size for gradle

This PR does nothing with logic and functionality, just conversion.

Motivation is that I'd like to improve SyncService in the future and I thought it might be a good idea to convert this class into kotlin first to make things fancier :)

Also, I had to increase Heap size for gradle, because it started to fail with OutOfMemory error.


bemyak:
This also closes #625 even without latest commit

nevenz:
> Increase heap size for gradle

I've recently removed this (77373a9e) from project's `./gradle.properties` and moved it to my `~/.gradle/gradle.properties`, thinking it depends on user's machine. But perhaps it depends more on the project size and needs to be included after all?

> This PR does nothing with logic and functionality, just conversion.
> This also closes #625 even without latest commit since I'm able to trigger sync with a command that issue's author mentioned

I'm not sure I understand how are these two issues connected?

app/src/main/java/com/orgzly/android/sync/SyncService.kt
```diff
@@ -0,0 +1,544 @@
...
+
+    private fun isTriggeredAutomatically(intent: Intent?): Boolean {
+        return intent != null && intent.getBooleanExtra(AppIntent.EXTRA_IS_AUTOMATIC, false)
+    }
+
+    private fun start(isTriggeredAutomatically: Boolean) {
+        if (BuildConfig.LOG_DEBUG) LogUtils.d(TAG)
+
+        syncTask = SyncTask()
+        syncTask!!.execute(isTriggeredAutomatically)
```
nevenz: We need to go through generated code after automatic conversion due to stuff like this. It's not perfect.

I'd use:
```
        syncTask = SyncTask().apply {
            execute(isTriggeredAutomatically)
       }
```

here for example.

app/src/main/java/com/orgzly/android/sync/SyncService.kt
```diff
@@ -0,0 +1,544 @@
...
+    }
+
+    private fun stop() {
+        if (BuildConfig.LOG_DEBUG) LogUtils.d(TAG)
+
+        syncStatus.set(SyncStatus.Type.CANCELING, null, syncStatus.currentBook, syncStatus.totalBooks)
+        announceActiveSyncStatus()
+
+        syncTask!!.cancel(false)
```
nevenz: This could be `syncTask?.cancel(false)`.

nevenz:
I've added few comments, there are more cases of `!!` which could be removed for example, there's one `TODO()` etc.

I usually take the opportunity when converting Java code to Kotlin to go through it and cleanup some stuff, we shouldn't be just relying on automatic conversion 100%.

Thanks!

bemyak:
Yes, sure. The point of this PR is to make further changes more clear, so they can be reviewed properly and not get lost in a complete file rewrite. I fixed some of the most nasty examples of not-so-good code generation, but obviously also missed some places.

> I'm not sure I understand how are these two issues connected?

My guess on the reason why the sync failed with issue author's command is NullPointer somewhere, because of missing function parameters. Kotlin is much better in null handling so probably that auto-generated code makes additions checks somewhere. The point is that now the sync can be triggered from CLI, or least That Works On My Machine :)

bemyak:
I haven't try it, but obviously it didn't work for https://github.com/orgzly/orgzly-android/issues/625 's author

nevenz:
Have you ran tests? I'm getting few of them failing, such as:
* testLoadRook
* testRepoAndShelfSetup
* testGetBooksFromAllRepos
* testRetrievingBook

BTW, if you can avoid merging back master to the working branch that would be great, to keep git log cleaner. Thanks!

bemyak:
Thank you for addressing this issues.

I fixed the problem that caused test failures (`null` handling in `getBooksFromAllRepos`), squashed all commits and rebased on top of current master.

nevenz:
> I fixed the problem that caused test failures (null handling in getBooksFromAllRepos), squashed all commits and rebased on top of current master.

Great, thank you. I just had to update it for 784290fc I did for v1.8.1.

PR #643, opened 2019-12-03
Git sync fixes

lfxgroove:
Should fix the force load and force save crashes that were seen in #543. I also moved the creation of a view a bit earlier, otherwise the Git sync view crashed on initialization.

Also updated to allow rudimentary HTTP support. Username and password is currently saved in clear-text, this is something we probably want to change?

The remaining items now are:

* [x]  Move git package to sync/ or repos/ (for now)
Does this need to happen? The current layout seems reasonable to me

* [x] Update JGit version to a newer one and solve the entrySet() problem
I don't think this can be solved without updating the android version we compile against which I guess we don't want to to keep backwards compatibility? Changing to a newer version yields the following error at runtime when trying to clone a repo:
```
2020-02-11 10:36:13.985 14248-14348/com.orgzly E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #4
    Process: com.orgzly, PID: 14248
    java.lang.RuntimeException: An error occurred while executing doInBackground()
        at android.os.AsyncTask$3.done(AsyncTask.java:354)
        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
        at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
        at java.util.concurrent.FutureTask.run(FutureTask.java:271)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: java.lang.NoSuchMethodError: No virtual method keySet()Ljava/util/concurrent/ConcurrentHashMap$KeySetView; in class Ljava/util/concurrent/ConcurrentHashMap; or its super classes (declaration of 'java.util.concurrent.ConcurrentHashMap' appears in /system/framework/core-oj.jar)
        at org.eclipse.jgit.attributes.FilterCommandRegistry.getRegisteredFilterCommands(FilterCommandRegistry.java:118)
        at org.eclipse.jgit.treewalk.TreeWalk.<init>(TreeWalk.java:375)
        at org.eclipse.jgit.treewalk.TreeWalk.<init>(TreeWalk.java:339)
        at org.eclipse.jgit.submodule.SubmoduleWalk.<init>(SubmoduleWalk.java:409)
        at org.eclipse.jgit.api.FetchCommand.fetchSubmodules(FetchCommand.java:167)
        at org.eclipse.jgit.api.FetchCommand.call(FetchCommand.java:247)
        at org.eclipse.jgit.api.CloneCommand.fetch(CloneCommand.java:305)
        at org.eclipse.jgit.api.CloneCommand.call(CloneCommand.java:204)
        at com.orgzly.android.repos.GitRepo.cloneRepo(GitRepo.java:149)
        at com.orgzly.android.repos.GitRepo.ensureRepositoryExists(GitRepo.java:95)
        at com.orgzly.android.repos.GitRepo.ensureRepositoryExists(GitRepo.java:85)
        at com.orgzly.android.ui.repo.git.GitRepoActivity$RepoCloneTask.doInBackground(GitRepoActivity.kt:447)
        at com.orgzly.android.ui.repo.git.GitRepoActivity$RepoCloneTask.doInBackground(GitRepoActivity.kt:436)
        at android.os.AsyncTask$2.call(AsyncTask.java:333)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245) 
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167) 
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641) 
        at java.lang.Thread.run(Thread.java:764) 
```

**Found during testing/changes from @amberin**
* [ ] Changing the storage location for a repo makes the repo unusable
What behaviour do we want here? Should we just copy the repo to the new location or not allow location changing at all for the first release of the feature?

* [x] mtime correction
I'm not sure what this is related to, could you fill med in @amberin?

* [ ] Change how conflict resolution is handled, push the conflicting merge branch to remote to allow fixing it somewhere else and then allow "force-pulling" a notebook. This probably doesn't fit in this PR but I wanted to remember it.


amberin:
The mtime issue was fixed in #543; "latest change" has since then shown the right time stamp.

amberin:
>     * Changing the storage location for a repo makes the repo unusable

I'm not sure what this refers to... Are you thinking of 935200c635b335953edef17de871549e5b3d8ab8 in my PR? That concerned *any* update to an existing repo. I found that (at least back then) no changes could be saved, because the repo was never properly identified.

If you are referring to this https://github.com/orgzly/orgzly-android/pull/543#discussion_r288986488 then I don't have an opinion. (Except that I still think it would be nicer to use private storage, remove the local path selection altogether and create a folder named after repo ID. But that would entail some more work, as previously mentioned.)

amberin:
Very happy to see some new commits on my favorite app! I wish I didn't rely on it so much in my daily life, then I could help out more with debugging...

amberin:
When cloning my repository with this build, the notebooks are imported, but don't have any links set. I have to manually select each notebook, choose "set link" and choose my (only) repository.

After that, I also have to force load each notebook before it gets any content. This is also very cumbersome if you have a large number of notebooks.

I don't remember this being the case previously, but it's possible that I just forgot.

amberin:
I have done some testing of this build primarily focusing on merging and conflict resolution. This is because I noticed a few months ago that I had actually lost data while using the build from my own branch.

From what I can tell, conflict-free merging works very smoothly.

Merging with conflicts also sort of works, but it's far from user-friendly. The new branch gets pushed from Orgzly to the remote. You can then fetch it on a computer and merge it into master, resolving the conflicts manually, commit and push master to remote. This is great, at least this is what I want. Upon the next Orgzly sync, the latest commit from master is loaded.

The problem is that until the branches have been merged, Orgzly is unable to sync any notebooks. Sync attempts result in the error message "Sync: Failed to merge <commit hash>".

In a way, this is of course great, because it forces the user to resolve the merge conflict. But I am afraid the tiny status text "Both local and remote notebook have been modified" is not enough to make the user understand what needs to be done. (That message doesn't even indicate that there was a conflict!) 

I also suspect that users may want the option of ignoring the conflict, and just moving on, depending on the importance of the notebook contents. After all, the conflict resolution and merging does require some Git knowledge that not all users possess.

Actually, the more I think about it, the more I believe I'd prefer a forced-merge approach. The way I see it, the beauty of Git is that you never really lose work as long as it has been committed at some point. I would probably prefer if Orgzly could force the merge for me (doing something like "git merge -X ours") and then alert me that a conflict was detected, and that I should take a look at the merge diff to make sure that I didn't lose anything important.

I'd love to hear @IvanMalison's opinion on this last point.

amberin:
I just found another issue. I haven't seen this before, but in the repo where I have been testing merging with conflicts, Orgzly is now always committing changes and new files to it's own branch, ie not to master.

IvanMalison:
> Actually, the more I think about it, the more I am in favor of a forced-merge approach. The way I see it, the beauty of Git is that you never irretrievably lose work as long as it has been committed at some point. I would probably prefer it if Orgzly could just force the merge for me (doing something like `git merge -X ours`) and then alert me that a conflict was detected, and that I should take a look at the merge diff to make sure that I didn't lose anything important.

Personally I would probably prefer no force merge behavior. While it is true that you never really lose anything with git, it would be easy to add tons of (not very well described) commits to the orgzly repo if you are using auto sync/commit on the orgzly and emacs sides. This would still make it easy to effectively lose information because it might not be obvious that you need to look back.

All that said, I don't think it would be hard to allow this to be configurable.

amberin:
I agree that making it configurable seems reasonable. Either way, a more informative notification is necessary.

I'm afraid that the last issue that I found is pretty major. It means that once a merge conflict has happened, even after a successful resolution and merge, Orgzly is loading notebooks from master, but committing changes to its "merge conflict branch". Syncing now appears to work fine, but Orgzly's commits never show up on other devices until they check out Orgzly's branch.

I will try to do some debugging, if I can find the time.

IvanMalison:
> I'm afraid that the last issue that I found is pretty major. It means that once a merge conflict has happened, even after a successful resolution and merge, Orgzly is loading notebooks from master, but committing changes to its "merge conflict branch"

Yeah, there just needs to be a way to set the "tracking branch"

nevenz:
> Should fix the force load and force save crashes that were seen in #543. I also moved the creation of a view a bit earlier, otherwise the Git sync view crashed on initialization.
> 
> Also updated to allow rudimentary HTTP support. Username and password is currently saved in clear-text, this is something we probably want to change?

As long as it's stored as a preference or anywhere else within the app's private directories, I think it's fine (as much as any other sensitive data is on Android). I don't know how we'd decrypt the password safely without asking the user for anything.

> * [ ]   Move git package to sync/ or repos/ (for now)
>   Does this need to happen? The current layout seems reasonable to me

No, I'll probably change my mind about it few more times anyway.

> * [ ]  Update JGit version to a newer one and solve the entrySet() problem
>   I don't think this can be solved without updating the android version we compile against which I guess we don't want to to keep backwards compatibility?

I forgot and can't find what was the issue there. It does compile with `5.6.0.201912101111-r` for example, but perhaps the issue is when running later?

> * [ ]  Changing the storage location for a repo makes the repo unusable
>   What behaviour do we want here? Should we just copy the repo to the new location or not allow location changing at all for the first release of the feature?
> Except that I still think it would be nicer to use private storage, remove the local path selection altogether and create a folder named after repo ID. 

I think using private storage is a good idea if it would make things simpler, as this has already become quite complex it seems. Also, once we start targetting Android 10 at one point, there could be issues with introduced privacy changes (scoped storage).

I do think user should be able to change the directory of an existing repository, which would require moving it I guess.

ScottFreeCode:
> > Also updated to allow rudimentary HTTP support. Username and password is currently saved in clear-text, this is something we probably want to change?

> As long as it's stored as a preference or anywhere else within the app's private directories, I think it's fine (as much as any other sensitive data is on Android). I don't know how we'd decrypt the password safely without asking the user for anything.

I'd be interested in how GitHub uses API tokens for HTTPS access. Seems like that would be preferable in terms of being able to limit permissions on the access through the token and revoke a token without having to change the user password.

lfxgroove:
> 
> 
> > Should fix the force load and force save crashes that were seen in #543. I also moved the creation of a view a bit earlier, otherwise the Git sync view crashed on initialization.
> > Also updated to allow rudimentary HTTP support. Username and password is currently saved in clear-text, this is something we probably want to change?
> 
> As long as it's stored as a preference or anywhere else within the app's private directories, I think it's fine (as much as any other sensitive data is on Android). I don't know how we'd decrypt the password safely without asking the user for anything.
> 
This is currently the case so I'll leave it as that.

> > * [ ]    Move git package to sync/ or repos/ (for now)
> >   Does this need to happen? The current layout seems reasonable to me
> 
> No, I'll probably change my mind about it few more times anyway.
> 
Marking as solved.

> > * [ ]   Update JGit version to a newer one and solve the entrySet() problem
> >   I don't think this can be solved without updating the android version we compile against which I guess we don't want to to keep backwards compatibility?
> 
> I forgot and can't find what was the issue there. It does compile with `5.6.0.201912101111-r` for example, but perhaps the issue is when running later?
> 
I tried running with this and get the entrySet() error I've mentioned in the other PR, see the full error at the top of this PR. Marking as solved for now I guess.

> > * [ ]   Changing the storage location for a repo makes the repo unusable
> >   What behaviour do we want here? Should we just copy the repo to the new location or not allow location changing at all for the first release of the feature?
> >   Except that I still think it would be nicer to use private storage, remove the local path selection altogether and create a folder named after repo ID.
> 
> I think using private storage is a good idea if it would make things simpler, as this has already become quite complex it seems. Also, once we start targetting Android 10 at one point, there could be issues with introduced privacy changes (scoped storage).
> 
> I do think user should be able to change the directory of an existing repository, which would require moving it I guess.
The current solution of selecting a directory works as far as I can tell, to me it would be great if we could let moving the git directory around be a later feature, i.e. I'll disable the button to change location after the initial clone. I'm also unsure of where we would place the files, since internal/private storage could have issues with how big the repo is. 

@amberin It seems I'm seeing the same issues. I found a null pointer exception in `SyncService.kt:handleTwoWaySync` where `currentRook` seems to be null for some reason. Do you have any insights as to why this might be? I'm not sure what I could have changed to introduce this behaviour.

lfxgroove:
Hi again, it's been some time since I've worked on this and unfortunately I don't think I'll have the time and motivation to finish this up right now, sorry for not letting you know earlier. The webdav implementation available is enough for my needs as far as I can tell and I'll use that instead. If anyone wants to continue the PR, feel free to!

nevenz:
@lfxgroove, thanks for all the work. Can this PR be merged perhaps and new ones created later (as we did before), if it will make things easier?

lfxgroove:
Sounds fine by me, i'm not sure of the state of the commit relating to master. The latest issues found were those discussed most recently. I.e, the syncing does not go as planned according to the text describing each book.

Last i tested was about a month back, i don't recall the code impacting any areas outside of the git handling so hopefully it is fine to merge.

amberin:
@lfxgroove:
> I found a null pointer exception in `SyncService.kt:handleTwoWaySync` where `currentRook` seems to be null for some reason. Do you have any insights as to why this might be? I'm not sure what I could have changed to introduce this behaviour.

It looks to me like the conversion of SyncService to Kotlin ended up using !! very liberally... Before the conversion, syncNamesake() returned a BookAction, but now it returns a nullable BookAction. Perhaps there is a good reason, but I can't see it. Since dataRepository.setBookLastActionAndSyncStatus() requires a BookAction argument, a "action!!" variable is passed to it, which unfortunately ends up being null when a new notebook appears in the Git repo.

amberin:
I believe I have found the cause of the NPE when importing new notebooks:

https://github.com/orgzly/orgzly-android/blob/master/app/src/main/java/com/orgzly/android/sync/SyncService.kt#L529

Before the Kotlin conversion (I tried an older branch), `handleTwoWaySync()` would call `GitRepo.syncBook()` with `currentRook` being null, and the logic in `syncBook()` would handle it (`syncBackNeeded` would get set). In the Kotlin version, the `currentRook!!` argument triggers an NPE when `currentRook` is null, so `syncBook()` never runs.

I have tried removing the !! and allowing a null VersionedRook as input to syncBook(), and it solved the issue. Should I open a separate PR, @nevenz ? Considering it's not strictly related to Git repos.

My build works fine with local and Git repos after these changes, but I don't have a good way to test WebDAV repos.

amberin:
> > I'm afraid that the last issue that I found is pretty major. It means that once a merge conflict has happened, even after a successful resolution and merge, Orgzly is loading notebooks from master, but committing changes to its "merge conflict branch"
> 
> Yeah, there just needs to be a way to set the "tracking branch" and the currently checked out branch.

@IvanMalison, I am working on fixing the current bugs around merging and conflict resolution. But I have become unsure of what behavior you had intended in the event of a merge conflict. Currently, the repo cannot be synced at all until the conflict has been resolved on another machine. Was this your intention?

I had assumed that your intention was that we should be able to keep syncing and committing to a temporary branch, and then later switch back to master after a successful merge. But in that case, how will Orgzly know when it is safe to do so?

IvanMalison:
> @IvanMalison, I am working on fixing the current bugs around merging and conflict resolution. But I have become unsure of what behavior you had intended in the event of a merge conflict. Currently, the repo cannot be synced at all until the conflict has been resolved on another machine. Was this your intention?

@amberin I mean, there is no real mechanism for resolving conflicts that is available in orgzly. How would we resolve conflicts without doing it on another machine?

IvanMalison:
@amberin Can you clarify what the behavior you are expecting is?

amberin:
@IvanMalison In the case of a merge conflict, stay on the temporary branch, but allow further syncing with new commits on the temporary branch.

I'll make a PR soon with a suggestion to achieve this behaviour, and automatic merging back into the main branch as soon as it's possible.

amberin:
> Sounds fine by me, i'm not sure of the state of the commit relating to master. The latest issues found were those discussed most recently. I.e, the syncing does not go as planned according to the text describing each book.
> 
> Last i tested was about a month back, i don't recall the code impacting any areas outside of the git handling so hopefully it is fine to merge.

I agree, this should be merged so that new PRs can be created on top of this work. This branch has worked very well for me (although I don't use HTTPS).

PR #657, opened 2020-01-17
Add first draft of "Show diff"(Help me pls!)

### LINKED ISSUE #633: Sync diff
- [X] I have searched for existing issues that may be the same as or related to mine.

Is it possible to add something like "view merge confilits diff"? I often fix merge conflicts by hand and for that I will have to skim through an entire file. Even the worst diff would be so much more helpful! Thank you for the amazing app btw!

[Original](https://github.com/orgzly/orgzly-android/issues/35#issuecomment-552599079).

Asalle:
Fixes #633 

Hello,
i wanted to add a simple window that shows diff between local notebook and remote one if a merge confict occurs (see issue). This is an early draft of what could it look like. I could find a way to display diff inside a window, could somebody help me with this?


nevenz:
I don't think this should be git-specific, but available for any repository type. That would require keeping the synced files around (which is probably a good idea anyway) and using them to generate the diff. Maybe using something like https://github.com/java-diff-utils/java-diff-utils ?

Asalle:
Thank you for the feedback, I'll review it and will get back to you.

Asalle:
@nevenz that is true, i should implement it for all repos, not just git, I myself don't use git :sweat_smile: 
Can anybody provide any mentoring on this? Right now I'm planning to do this:
1. Add function `StoreBooks` to `SyncRepo` interface and use it to download all remote books to some local folder
2. Find a place where the sync happens (working on this rn) and instead of syncing with remote, fetch the remote notebooks first and then sync them. If any merge conficts appear - show diff.

nevenz:
I don't think there's a need for changing the interface. As you suggest in <span>#</span>2, we can simply keep already used notebooks files around. See the creation and deletion of temporary files in `loadBookFromRepo` and `saveBookToRepo`.

Asalle:
@nevenz on latest upsteam/master I cannot sync with dropbox, only `webdav` and `directory`, am I doing something wrong?

nevenz:
> on latest upsteam/master I cannot sync with dropbox, only webdav and directory, am I doing something wrong?

Check the build variant you're using.  Perhaps it's one of the `fdroid*` ?

Asalle:
yup, that was the reason. Thanks!

yamsu:
This would be a really useful addition to the feature set. Syncing issues are very common when using dropbox integration, as sync has to be manually activated.

Looking forward to this merge.

Asalle:
damn, I don't remember closing this!

Asalle:
damn, I might have accidentally overwritten my master branch (who even does work on master, jeesh!)
Trying to work on it, living ALL THESE YEARS without diff is hard... Any help welcome!

berteh:
+1
this feature would be most useful. thanks for working on this!

PR #661, opened 2020-01-19
Add preference to set default reminder time

### LINKED ISSUE #651: Reminders at midnight
https://github.com/orgzly/orgzly-android/blob/a7d2a4ddec6f2fa974e8aaceb8c7ac213088c801/app/src/main/java/com/orgzly/android/reminders/ReminderService.kt#L396

Is this line responsible for the fact that notes with only a day set, have their reminders at 9:00 AM?
If so, wouldn't it be more logical to make that 00:00 AM? (If I am not mistaken, that was the original behavior).

Context: I usually do not set a time for my notes, so I expect them to either:
* Not have a reminder (if want a reminder, I will set a time);
* Or a reminder at midnight.

The thing is, my phone is in DND mode at midnight, so reminders at midnight don't disturb me. My phone stops DND mode at 7:00 AM, so the rather useless reminders at 9:00 AM do disturb me.
You might argue that I could set my phone to hold DND mode till 9:00 AM, but then notes that do have a time (the urgent ones, for me) won't come thru. (I never have reminders before 7:00 AM, so till that time DND mode is fine.)

andipabst:
This fixes #651.

The default time for this preference is still 9 o'clock, like it was before. But now a user is able to change this time.

### DIFF
--- a/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
+++ b/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
@@ -311,6 +311,17 @@ public class AppPreferences {
                 context.getResources().getString(R.string.pref_default_snooze_type));
     }
 
+    public static int reminderDailyTime(Context context) {
+        String key = context.getResources().getString(R.string.pref_key_daily_reminder_time);
+        return getStateSharedPreferences(context).getInt(key,
+                context.getResources().getInteger(R.integer.pref_default_daily_reminder_time));
+    }
+
+    public static void reminderDailyTime(Context context, int value) {
+        String key = context.getResources().getString(R.string.pref_key_daily_reminder_time);
+        getStateSharedPreferences(context).edit().putInt(key, value).apply();
+    }
+
     public static boolean showSyncNotifications(Context context) {
         return getDefaultSharedPreferences(context).getBoolean(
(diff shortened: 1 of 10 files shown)


nevenz:
Thank you!

Few things:

- [ ] `onSharedPreferenceChanged` is not called, so reminders are not rescheduled after preference is changed
- [ ] Preference is not disabled when there are no reminders enabled (also in `SettingsFragment`)
- [ ] Can we rename resources to `pref_key_daily_reminder_time` and `pref_default_daily_reminder_time` ?
- [ ] There are some deprecated methods called in `TimePreferenceFragment`
- [ ] Translations are overwritten by those on https://crowdin.com/project/orgzly, so you can remove it
- [ ] Dialog's title lacks some bottom padding

andipabst:
Thanks for your review. I made the changes you suggested.

> There are some deprecated methods called in `TimePreferenceFragment`

They are deprecated from Android M (API version 23) on, so I wrapped the calls in an if-clause.

nevenz:
> The problem is that onSharedPreferenceChanged is not called in the first place.

This was due to "state" SharedPreferences used, instead of the default one.

Also fixed `timePicker.currentHour = minutes` and moved the preference up in Settings.

Thanks!

PR #692, opened 2020-04-21
Recognize + as valid bullet symbol for unordered lists

### LINKED ISSUE #558: Render list markup as bullet points
It would be nice to have an option to enable the rendering of `-` and `+` lists as bullet points.

I'm not sure this should be mandatory (some users may prefer more plain-text style lists), so an option in the settings to enable this would be great.

Possibly related: #469

debanjum:
**Description**: 
Make Orgzly recognize `+` as a valid bullet symbol for unordered lists  
   
**Implemented**:
  * [X] Add `+` as valid bullet symbols for unordered list items and plain list items with checkboxes

**References**: 
  * [Org-Mode: Plain Lists](https://www.gnu.org/software/emacs/manual/html_node/org/Plain-lists.html)
  * 8e0493725b116d34004ba8dbd454bd49cdd5d033: Recognizing `-` as valid bullet for plain list item

Resolves #558

### DIFF
--- a/app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
+++ b/app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
@@ -36,7 +36,9 @@ class EditTextWithMarkup : AppCompatEditText {
         // TODO: Remove space at the end of bullet
         private val listItemTypes = arrayOf(
                 ListItem(Pattern.compile("^(\\s*)-\\s+\\[[ X]](.*)"), "- [ ] "),
-                ListItem(Pattern.compile("^(\\s*)-(.*)"), "- ")
+                ListItem(Pattern.compile("^(\\s*)-(.*)"), "- "),
+                ListItem(Pattern.compile("^(\\s*)\\+\\s+\\[[ X]](.*)"), "+ [ ] "),
+                ListItem(Pattern.compile("^(\\s*)\\+(.*)"), "+ ")
         )
 
         private var currentListItem : CurrentListItem? = null
--- a/app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
+++ b/app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
@@ -51,7 +51,8 @@ object OrgFormatter {
 
     private val LOGBOOK_DRAWER_PATTERN = drawerPattern(LOGBOOK_DRAWER_NAME)
 
-    private val CHECKBOXES_PATTERN = Pattern.compile("""^\s*-\s+(\[[ X]])""", Pattern.MULTILINE)
+    private const val PLAIN_LIST_CHARS = "-\\+"
+    private val CHECKBOXES_PATTERN = Pattern.compile("""^\s*[$PLAIN_LIST_CHARS]\s+(\[[ X]])""", Pattern.MULTILINE)
 
     private const val FLAGS = Spanned.SPAN_EXCLUSIVE_EXCLUSIVE
 


app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
```diff
@@ -36,7 +36,11 @@ class EditTextWithMarkup : AppCompatEditText {
         // TODO: Remove space at the end of bullet
         private val listItemTypes = arrayOf(
                 ListItem(Pattern.compile("^(\\s*)-\\s+\\[[ X]](.*)"), "- [ ] "),
-                ListItem(Pattern.compile("^(\\s*)-(.*)"), "- ")
+                ListItem(Pattern.compile("^(\\s*)-(.*)"), "- "),
+                ListItem(Pattern.compile("^(\\s*)\\+\\s+\\[[ X]](.*)"), "+ [ ] "),
+                ListItem(Pattern.compile("^(\\s*)\\+(.*)"), "+ ")
+                ListItem(Pattern.compile("^(\\s*)\\*\\s+\\[[ X]](.*)"), "* [ ] "),
+                ListItem(Pattern.compile("^(\\s*)\\*(.*)"), "* ")
```
GuiltyDolphin, line 43: I think Org only supports asterisks if they are indented, so we might need e.g.,

```
                 ListItem(Pattern.compile("^(\\s+)\\*(.*)"), "* ")
```

app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
```diff
@@ -36,7 +36,11 @@ class EditTextWithMarkup : AppCompatEditText {
         // TODO: Remove space at the end of bullet
         private val listItemTypes = arrayOf(
                 ListItem(Pattern.compile("^(\\s*)-\\s+\\[[ X]](.*)"), "- [ ] "),
-                ListItem(Pattern.compile("^(\\s*)-(.*)"), "- ")
+                ListItem(Pattern.compile("^(\\s*)-(.*)"), "- "),
+                ListItem(Pattern.compile("^(\\s*)\\+\\s+\\[[ X]](.*)"), "+ [ ] "),
+                ListItem(Pattern.compile("^(\\s*)\\+(.*)"), "+ ")
+                ListItem(Pattern.compile("^(\\s*)\\*\\s+\\[[ X]](.*)"), "* [ ] "),
```
GuiltyDolphin, line 42: See my below comment about asterisks, so here we'd probably want (and apologies if this is incorrect):

```
                ListItem(Pattern.compile("^(\\s+)\\*\\s+\\[[ X]](.*)"), "* [ ] "),
```

app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
```diff
@@ -51,7 +51,8 @@ object OrgFormatter {
 
     private val LOGBOOK_DRAWER_PATTERN = drawerPattern(LOGBOOK_DRAWER_NAME)
 
-    private val CHECKBOXES_PATTERN = Pattern.compile("""^\s*-\s+(\[[ X]])""", Pattern.MULTILINE)
+    private const val PLAIN_LIST_CHARS = "-\\+\\*"
+    private val CHECKBOXES_PATTERN = Pattern.compile("""^\s*[$PLAIN_LIST_CHARS]\s+(\[[ X]])""", Pattern.MULTILINE)
```
GuiltyDolphin, line 55: Would this account for the case with asterisks needing to be indented? If not, we might need to provide an OR here, with the asterisk requiring spacing prior to the character.

ScottFreeCode:
As far as I know Orgzly adds space before asterisks at the start of lines in a note's content automatically, and works correctly for indented lists; it shouldn't need any special handling, unless I'm missing something.

GuiltyDolphin:
@ScottFreeCode ah, I did not know this to be the case! This should be fine then :)

debanjum:
> As far as I know Orgzly adds space before asterisks at the start of lines in a note's content automatically, and works correctly for indented lists; it shouldn't need any special handling, unless I'm missing something.

Yeah, that's what it was seeming like while testing these changes on Android.

Thanks for the review comments and clarifications @GuiltyDolphin and @ScottFreeCode

app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
debanjum: Resolving thread based on [discussion](https://github.com/orgzly/orgzly-android/pull/692#issuecomment-620202416)

app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
debanjum: Resolving conversation based on [discussion](https://github.com/orgzly/orgzly-android/pull/692#issuecomment-620202416)

app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
GuiltyDolphin, line 43: If possible, could we refactor these regular expressions, as we have a single format for each of the list and checklist regexes in which only the bullet character changes.

Not essential, but would make any future changes easier if we needed to change how lists are parsed for all bullets.

app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
```diff
@@ -36,7 +36,11 @@ class EditTextWithMarkup : AppCompatEditText {
         // TODO: Remove space at the end of bullet
         private val listItemTypes = arrayOf(
                 ListItem(Pattern.compile("^(\\s*)-\\s+\\[[ X]](.*)"), "- [ ] "),
-                ListItem(Pattern.compile("^(\\s*)-(.*)"), "- ")
+                ListItem(Pattern.compile("^(\\s*)-(.*)"), "- "),
+                ListItem(Pattern.compile("^(\\s*)\\+\\s+\\[[ X]](.*)"), "+ [ ] "),
+                ListItem(Pattern.compile("^(\\s*)\\+(.*)"), "+ ")
```
nevenz, line 41: This doesn't compile -- there's a `,` missing at the end.

nevenz:
I'm not sure that supporting `*`-lists should be done. Since this is about adding new items, I'd rather not encourage the use of that type of list.

Also, item with `*` in the first column is prepended with space, but sub-items are not touched, so indentation becomes off.

GuiltyDolphin:
> I'm not sure that supporting `*`-lists should be done. Since this is about adding new items, I'd rather not encourage the use of that type of list.
> 
> Also, item with `*` in the first column is prepended with space, but sub-items are not touched, so indentation becomes off.

Yeah... I agree. I personally only use `-` and `+` bullets, and I worry that introducing `*` could a) cause a bit of confusion for those learning the syntax, and b) potentially introduce some bugs now and/or down the line (sounds like we could already experience some issues with that indentation?).

@debanjum would you like to remove support for `*` as a bullet item, and we can stick to just having `+` and `-`?

app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
debanjum: Fixed on removing support for * symbol

debanjum:
> @debanjum would you like to remove support for `*` as a bullet item, and we can stick to just having `+` and `-`?
Removed support for `*` as bullet item in edit mode.

app/src/main/java/com/orgzly/android/ui/views/EditTextWithMarkup.kt
debanjum: I'm new to Kotlin and Android development in general. Would like to avoid mixing refactoring with the minimal feature update! :slightly_smiling_face:

app/src/main/java/com/orgzly/android/util/OrgFormatter.kt
```diff
@@ -51,7 +51,8 @@ object OrgFormatter {
 
     private val LOGBOOK_DRAWER_PATTERN = drawerPattern(LOGBOOK_DRAWER_NAME)
 
-    private val CHECKBOXES_PATTERN = Pattern.compile("""^\s*-\s+(\[[ X]])""", Pattern.MULTILINE)
+    private const val PLAIN_LIST_CHARS = "-\\+\\*"
```
GuiltyDolphin, line 54: I think you can remove the `*` from here.

nevenz:
Cool. Can we get a rebase and a single commit for this? As changes were reverted, so log is confusing. Thanks!

debanjum:
> Cool. Can we get a rebase and a single commit for this? As changes were reverted, so log is confusing. Thanks!

Rebased the change into a single commit. Hope this looks fine?

GuiltyDolphin:
Looks good :+1:

nevenz:
Thanks!

Issue #565, opened 2019-06-25
Make internal links clickable in widget/headers


GuiltyDolphin:
Currently, internal links in headers aren't clickable when viewed from the notes widget, and there is no way to click the link from the note view itself (only from view with multiple notes).

It would be really useful if it were possible to follow internal links in headings from the widget view, as well as all links in headings from the view mode of individual links.

Related: #181

Asalle:
possible duplicate of #181 ?

GuiltyDolphin:
@Asalle I'm specifically referring to _internal links_ (i.e., to other headings) not displaying correctly in the notes widget view, rather than during note editing. I've offhandedly mentioned the other issue here as an addition, apologies if that made the request confusing!

PR #702, opened 2020-05-03
Support relative links to headers inside notebooks

### LINKED ISSUE #694: Support links to headers inside notebooks
- [X] I have searched for existing issues that may be the same as or related to mine.

I have a notebook `my-notebook.org` with following content:
```
* MyHeading
```

I want to have a link to `MyHeading` in my other notebook, `second-notebook.org`, so I put this inside it:
```
* Unrelated heading
```

Asalle:
Fixes #694 
Related to #565 

Now, if you click on the link "[[file:"notes.org::Heading1"][link]]" it opens the selected note (in "editing" mode).

Please help me patch up this frankenstein, I just learned about kotlin a week ago :)

### DIFF
--- a/app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
+++ b/app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
@@ -29,12 +29,12 @@ class LinkFindTarget(val path: String) : UseCase() {
         }
     }
 
-    private fun isAbsolute(path: String): Boolean {
-        return path.startsWith('/')
+    private fun isAbsolute(bookPath: String): Boolean {
+        return bookPath.startsWith('/')
     }
 
-    private fun isMaybeBook(path: String): BookName? {
-        val file = File(path)
+    private fun isMaybeBook(bookPath: String): BookName? {
+        val file = File(bookPath)
 
         return if (!hasParent(file) && BookName.isSupportedFormatFileName(file.name)) {
             BookName.fromFileName(file.name)
--- a/app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
+++ b/app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
@@ -3,6 +3,8 @@ package com.orgzly.android.usecase
 import android.os.Environment
 import com.orgzly.android.BookName
 import com.orgzly.android.data.DataRepository
+import com.orgzly.android.db.entity.Book
+import com.orgzly.android.db.entity.NoteView
 import java.io.File
 
 class LinkFindTarget(val path: String) : UseCase() {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
```diff
@@ -15,9 +17,23 @@ class LinkFindTarget(val path: String) : UseCase() {
     }
 
     private fun openLink(dataRepository: DataRepository, path: String): Any {
+        val regex = """(.*)\:\:(.*)""".toRegex()
         return if (isAbsolute(path)) {
             File(path)
-
+        } else if (regex.matches(path)) {
+            val matchResults = regex.matchEntire(path)
```
nevenz, line 24: I think we want to avoid matching twice.

app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
```diff
@@ -15,9 +17,23 @@ class LinkFindTarget(val path: String) : UseCase() {
     }
 
     private fun openLink(dataRepository: DataRepository, path: String): Any {
+        val regex = """(.*)\:\:(.*)""".toRegex()
```
nevenz, line 20: Perhaps `*` can be included in the regex. Something like:

```suggestion
        val regex = """(.*)::\*(.*)""".toRegex()
```
If not, we need to check for it below, having in mind support for other link types in the future, like:

```
file:sometextfile::NNN (jump to line number)
file:projects.org::some words (text search)
file:projects.org::#custom-id (headline search)
```

app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
```diff
@@ -15,9 +17,23 @@ class LinkFindTarget(val path: String) : UseCase() {
...
         return if (isAbsolute(path)) {
             File(path)
-
+        } else if (regex.matches(path)) {
+            val matchResults = regex.matchEntire(path)
+            val (_, notebook, heading) = matchResults!!.groupValues
```
nevenz, line 25: `!!` is generally avoided in Kotlin. A single match and checking for its result should do it.

app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
```diff
@@ -15,9 +17,23 @@ class LinkFindTarget(val path: String) : UseCase() {
...
+            isMaybeBook(notebook)?.let { bookName ->
+                dataRepository.getBook(bookName.name)?.let {
+                    val allNotes = dataRepository.getNotes(it.name)
+                    for (note in allNotes) {
```
nevenz, line 29: See `getNotesByTitle` in `DataRepository`. We have to avoid going through all book's notes.

I'd create a new `getNotesByTitle` that accepts book ID in addition to the title.

app/src/main/java/com/orgzly/android/usecase/LinkFindTarget.kt
```diff
@@ -29,12 +45,12 @@ class LinkFindTarget(val path: String) : UseCase() {
         }
     }
 
-    private fun isAbsolute(path: String): Boolean {
-        return path.startsWith('/')
+    private fun isAbsolute(bookPath: String): Boolean {
+        return bookPath.startsWith('/')
```
nevenz, line 49: This is not consistent with the name above. I don't care much about the name itself. Perhaps just `link` would make the most sense?

hahuang65:
Is there any further progress on this?

PR #713, opened 2020-05-18
Grant storage permission in tests to prevent BooksTest.testExport failing

artbristol:
The Android OS permissions dialog pops up the first time you run this test

According to the javadoc of `GrantPermissionRule`, "Once a permission is granted it will apply for all tests running in the current Instrumentation" so I added it to to the top-level test to avoid ambiguity when the test runner chooses a different test ordering


nevenz:
Great, thanks. I've been using a script for this so far.

PR #741, opened 2020-07-10
Add option to prepend new notes

### LINKED ISSUE #140: insert new note at beginning of notebook
I'm transitioning from Google Keep, where new notes appear at the top. I often make notes spontaneously, of which some then turn out to be irrelevant, so i don't really bother with cleaning them. In Keep that's no problem because they just slowly "sink" to the bottom as I add new ones, but in orgzly these irrelevant notes fill up the front place of my notebooks. Is it possible to add an option to Orgzly that creates a new note at the top? cheers!

drzisga:
Added option to prepend new notes at the beginning of a notebook. This should fix #140.


nevenz:
Thank you!

PR #760, opened 2020-09-22
Git push over SSH only once per sync

amberin:
The rationale behind this is that I have found Git syncing multiple notebooks over SSH to be very slow. I realized that separate SSH sessions were being initiated for each updated notebook.

I believe there is no reason to initiate more than two SSH sessions: one for fetch the repo, and one for pushing it after all commits have been made.


amberin:
ping @IvanMalison @lfxgroove

PR #768, opened 2020-10-13
Note WebDAV support in the README

bplaxco:
Noticed that the readme mentioned the other repo types but left out WebDAV.


nevenz:
Thanks!

Issue #496, opened 2019-03-16
Allow file links relative to repository root


memeplex:
Please, for local repositories this is very important since it allows to keep a consistent structure both in your computer and in orgzly. I think file:/xxx should be relative to your sdcard but file:xxx should be relative to your repo. I can't figure out why this isn't like that.

memeplex:
Implementing this feature will mostly solve the problem of notebooks nested in a hierarchy as discussed in https://github.com/orgzly/orgzly-android/issues/485.

I believe this feature is low cost and pays too much: by simply differentiating between `file:/xxx` and `file:xxx` you get:

1. Consistency between the links in your computer and the links in orgzly.

2. Consistency between the tree structure in your computer and the flattened structure in orgzly.

3. Ability to access arbitrary notebooks in the hierarchy.

It's true that this still requires to manually create index files to access the content buried in the hierarchy, but still it's a far cry from no being able to access this content at all!!!

memeplex:
I changed my mind regarding `file:xxx`. Despite the fact that currently "relative to repo" and "relative to file" mean the same thing, since orgzly just sees a list of files under the repo path, room should be made in advance for behavior more closely aligned with org mode, because hopefully one day `file:proj1/notes.org` would be accessible and then relative file links in `proj1/notes.org` would have to be accessed relative to `proj1` as in org mode. So I propose this behavior:

* `file:/xxx` relative to sdcard
* `file:xxx` relative to notebook, which currently is equivalent to relative to repo

nevenz:
> file:/xxx relative to sdcard

There could be multiple "external" storages. For example the main emulated one, but also removable SD card at the same time. These all currently work (after 1a63a685bc987a3e7693ff216eac380ba60e689c):

```
file:file-on-main-storage.txt
file:/sdcard/file-on-main-storage.txt
file:/storage/emulated/0/file-on-main-storage.txt

file:/storage/0F18-3901/file-on-sd-card.txt
```

I think an option is needed to map `/` to whatever user wants it to be. Maybe the same with relative paths - some users might want `file:xxx` to be relative to SD card, not the repository (which might not even exist, or is remote).

Ypot:
Where would fit this syntax?
[[file-on-main-storage.txt]]

nevenz:
> [[file-on-main-storage.txt]]

This is the same as `file:file-on-main-storage.txt`

vit1-irk:
[...]

PR #773, opened 2020-10-26
Add two settings to configure relative path for file links.

xiaoruoruo:
Addresses #496 

Current behavior was
- `file:/xxx` relative to root.
- `file:xxx` relative to Environment.getExternalStorageDirectory().getPath().

The new behavior in this commit is:
- `file:/xxx` relative to pref AppPreferences#fileAbsoluteRoot(), default to "/" 
- `file:xxx` relative to pref AppPreferences#fileRelativeRoot(), default to Environment.getExternalStorageDirectory().getPath()

Added two basic settings in the Image category for these two prefs. The setting is text only, could potentially be directoyr picker.

The path following behavior is changed in `ImageLoader` and `Widget.followLinkToFile`, now both uses `LinkFindTarget`.

### DIFF
--- a/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
+++ b/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
@@ -633,6 +633,25 @@ public class AppPreferences {
     }
 
     /*
+     * File relative path.
+     */
+
+    /** Root for file:/xxx links */
+    public static String fileAbsoluteRoot(Context context) {
+        return getDefaultSharedPreferences(context).getString(
+                context.getResources().getString(R.string.pref_key_file_absolute_root),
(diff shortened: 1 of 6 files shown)


nevenz:
I've moved (and renamed in the process) preferences under "Note", as they are not used only for images.

I've also removed default from `pref_key_file_absolute_root` for consistency (now "Not set", was "/"). I think it's obvious what default is for that one, much more then what it is for relative links. Ideally both would have it, but there's more involved in setting it for relative links as it's not static.

We might want to add a preference for `~/` as well in the future. And instead of current default for relative links use repository's location (per #496).

Thanks!

rjack:
Hi, this a great feature, just wanted to point out that the [documentation lists this option as already available](http://www.orgzly.com/help#a7a9f79) while it's actually yet to be released.

Issue #74, opened 2017-05-07
Add support for wide Org Tables


dustinswan:
If an Org Table is too wide, it will wrap making it difficult to read.

Would it be possible to disable wrapping and allow horizontal scrolling for Org Tables only?

vzaliva:
Because the phone screen is quite narrow, the tables with more than 2-3 columns do not fit horizontally. I think it is very important to show them without wrapping. Attached is a screenshot illustrating the problem:

![2018-07-16 09 37 27](https://user-images.githubusercontent.com/86581/42747324-4302ed32-88dc-11e8-9a57-9f4ff5a8b8be.jpg)

j3kyl:
Same here!

artbristol:
I would approach this as follows:
Change `NoteViewModel.content` to be a `List<NoteContent>` instead of a String. `NoteContent` would have subclasses for `Text` (a plain block of text) and `Table` (a table) and any other richly displayed content (e.g. Logbook drawer).
I'd also consider changing the DB entity `Note` to normalise the content into another table with a row for each item in the list.
Then in the UI, it should be easy enough to display tables differently from plain text.
Any thoughts @nevenz ?

nevenz:
I'm not sure what will be the best way to do this. Using Spans is another option.

Perhaps we should check how some other apps do it. There are libraries too - for example, I have Markwon in my notes as something to check.

artbristol:
Sure, having looked into the code a bit more, I think spans might be a better idea.
Perhaps an overlay span with fixed width, and a link to another view that allows (just) the table to be edited?

nevenz:
> Perhaps an overlay span with fixed width, and a link to another view that allows (just) the table to be edited?

Sure, sounds good. What would clicking of the link do? Are you thinking something like a popup window, entering edit mode for entire content, just that cell, or something else?

BTW, horizontal scrolling of the content is also needed (it's actually what the main issue here is - one could use monospaced font even now as a workaround for aligning, but still have the same problem with wrapping).

artbristol:
I was thinking of a new combined view/edit screen. The table's content could be truncated or collapsed somehow on the main note view (like a drawer) to avoid needing horizontal scrolling there. You'd need to go to the new screen to view the table.

decadent:
[...]

PR #799, opened 2020-12-31
Start formatting tables with horizontal scroll and fixed-width font

artbristol:
Followup from discussion on https://github.com/orgzly/orgzly-android/issues/74

In this approach the note's content is split into alternating table and text (i.e. non-table) content. The tables are then formatted with a fixed-width font, and horizontal scrolling is used to prevent text wrapping (wrapping makes tables basically unusable)

Before merging I'd at least need to address:

* test breakage
* adding more tests
* consider what to do about editing tables
* consider pushing down the parsing of table content into `org-java` library and storing result of the table/text parsing process in the database

### DIFF
--- a/app/src/androidTest/java/com/orgzly/android/espresso/ExternalLinksTest.kt
+++ b/app/src/androidTest/java/com/orgzly/android/espresso/ExternalLinksTest.kt
@@ -59,7 +59,7 @@ class ExternalLinksTest(private val param: Parameter) : OrgzlyTest() {
         onBook(0).perform(click())
 
         // Click on link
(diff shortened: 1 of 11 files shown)


artbristol:
@nevenz This works better than I expected. Do you think the ability to edit a table nicely is necessary or could we implement this display enhancement first?
Also this PR parses the tables on the UI thread which I thought would be bad, performance-wise, but it seems fine it me.

artbristol:
@nevenz I added a very bare-bones edit screen. I tried reusing the `NoteViewModel` but there is still a lot of duplication.
I was planning to add icons on the edit screen for actions like "reformat", "add/delete column", "add/delete row". What do you think?

Issue #804, opened 2021-01-10
Make search expression "hasOwnTag" to support "not" operator


LambdaAlpha:
- I have searched for existing issues that may be the same as or related to mine.

I don't know why this is not supported by "hasOwnTag" while it is supported by "hasTag".

In the file `SqliteQueryBuilder.kt`

```kt
is Condition.HasTag -> {
    repeat(2) { arguments.add("%${expr.tag}%") }
    not(expr.not, "(COALESCE(tags, '') LIKE ? OR COALESCE(inherited_tags, '') LIKE ?)")
}

is Condition.HasOwnTag -> {
    arguments.add("%${expr.tag}%")
    "tags LIKE ?"
}
```

If we change them to

```kt
is Condition.HasTag -> {
    repeat(2) { arguments.add("%${expr.tag}%") }
    not(expr.not, "(COALESCE(tags, '') LIKE ? OR COALESCE(inherited_tags, '') LIKE ?)")
}

is Condition.HasOwnTag -> {
    arguments.add("%${expr.tag}%")
    not(expr.not, "(COALESCE(tags, '') LIKE ?)")
}
```

Will it work?

I have some tags which I don't want to be inherited by subitems, i.e. some meta tags which indicating some states of the items. Sometimes I need to filter those items which don't have such a tag, `.t.tag` doesn't work in this situation, so I need `.tn.tag`.

nevenz:
> I don't know why this is not supported by "hasOwnTag" while it is supported by "hasTag".

No idea what was I thinking.

> If we change them to

Yeah, I think that should work. There are a few more places that need minor updates - `Condition.kt` and `DottedQueryParser.kt` for example. A simple test would be useful in `QueryTest.kt` too.

PR #805, opened 2021-01-11
support for .tn search operator

LambdaAlpha:
#804 

Signed-off-by: LambdaAlpha <2657922686@qq.com>


doak:
That seems to be quite useful: That way you could check if the tag was actually inherited.

Please don't forget to update the documentation as well :)

nevenz:
Thank you!

PR #843, opened 2021-03-09
Sticky Notifications initial feature commit.

akkord64:
Added an option to enable the reminders to be sticky and not swiped away until completed.


akkord64:
Found a small issue I want to resolve before commit.

PR #855, opened 2021-05-22
Per book encryption of notebooks

heinzelotto:
Hi,

in #687 @nevenz suggested per-book encryption. I had already hacked this together a long time ago and now I want to share it. Please try it out to see whether this is a good approach to build the full feature on.

Features:
* Global setting to set the default passphrase that can then be easily used for each individual notebook. This passphrase is also tried when downloading existing remote notebooks.
* Longtap a notebook and "Set Encryption" allows to set the passphrase for an individual notebook. When switching from unencrypted->encrypted there is then also the option to delete the existing unencrypted remote notebook. This is important because I currently reject the case where there are both "book.org" and "book.org.gpg" remote books present
* Database has been extended to store the encryption status for each notebook. TODO: database migration, I'm lacking experience here
* I added some cases for the sync & conflict logic, but this needs to be inspected much more rigorously. Test cases are needed here.


lifebarier:
Is this effort dead?

Issue #25, opened 2017-02-28
Add support for Google Drive synchronization


angrybacon:

medaka:
Is there somebody struggling for this feature?
I am investigating Google Drive apis and this repository and would like to share ideas if somebody is working for this.

c-okelly:
I would be interested in working on it as well! I don't think there is currently a timeline on it and I would very much like the feature!

rafaelleru:
I'm interested in work on this, any help about where to start looking? I mean some classes or packages that can be candidates to modify

nevenz:
Service-side - new class is needed implementing [Repo](https://github.com/orgzly/orgzly-android/blob/master/app/src/main/java/com/orgzly/android/repos/Repo.java) interface
UI-side - new fragment is needed to display directory/auth view

Those are tied together in [ReposActivity](https://github.com/orgzly/orgzly-android/blob/master/app/src/main/java/com/orgzly/android/ui/ReposActivity.java) and [RepoFactory](https://github.com/orgzly/orgzly-android/blob/master/app/src/main/java/com/orgzly/android/repos/RepoFactory.java).

Checking existing [DropboxRepo](https://github.com/orgzly/orgzly-android/blob/master/app/src/main/java/com/orgzly/android/repos/DropboxRepo.java) and [DirectoryRepo](https://github.com/orgzly/orgzly-android/blob/master/app/src/main/java/com/orgzly/android/repos/DirectoryRepo.java) is probably a good start.

(There's a lot of room to DRY things up, as this is all very similar. But I wouldn't worry about that right now if you're just starting - copying what `DropboxRepo` does and how is used is fine.)

relan:
Note that `com.google.android.gms:play-services-drive` is proprietary software. If it becomes a mandatory dependency, Orgzly won't be updated on F-Droid.

rafaelleru:
I will keep it in mind @relan and try to find a way to solve this problem.

rafaelleru:
Also I like to try to add nextcloud/ownclud integration appart from google drive. What do you think?

klapauciusisgreat:
@relan: isn't the dropbox integration proprietary as well ? Why is integration with drive separate ? 

[...]

PR #867, opened 2021-06-22
Google Drive Integration

ngharrison:
This implementation reflects what was done for syncing with Dropbox.

The code is there, but it's not working yet because the mDriveService becomes null before it is used to access the API. I believe this requires better knowledge of the app system that's already set up, so I'm starting the discussion here.


sjhuang26:
I personally would love a Google Drive sync feature and am looking to work on this issue.

According to the maintainer of another popular open-source note-taking app [1], Google policy requires an expensive "independent Google-approved audit" to use the necessary APIs.

However, we can use a scope like `https://www.googleapis.com/auth/drive.file` for which the policy doesn't apply (because the scope is non-restricted) and which should cover all of the user's needs. The app can "view and manage Google Drive files and folders that you have opened or created with this app" [3]. Note that "folders are treated as a type of file" [4].

Even if we needed a restricted scope, according to the official documentation [2], "for apps using restricted scopes, a restricted scope verification must be performed ... if you store restricted scope data on servers (or transmit), then you need to go through a security assessment" from "third-party vendors." Since we don't store user data, we don't need the security assessment.

@nevenz @ngharrison Would a third-party security assessment be required? IMO it's not.

[1] https://github.com/laurent22/joplin/issues/402#issuecomment-506737256
[2] https://developers.google.com/drive/api/v3/about-auth
[3] https://developers.google.com/identity/protocols/oauth2/scopes#drive
[4] https://developers.google.com/drive/api/v3/about-files

ngharrison:
Hey @sjhuang26, I'm not sure about the security assessment. You already know more about it than I do.

As for the code side, I'm not working on this anymore -- other priorities. I made a few more attempts after making this PR to fix the issues I was having and I could push those changes. From the current commit, the code builds, but any Drive functions fail because the Drive object becomes null. I tried moving the creation of that object into `GoogleDriveClient.java` from `GoogleDriveRepoActivity.kt`, but I couldn't get it to build.

One of my hopeful assumptions was that `GoogleSignIn` would handle the authentication tokens for me. I'm not sure if this was one of the problems, but it might have to be handled manually, how `DropboxClient.java` does it.

ngharrison:
Just added a commit of the most recent attempts.

ngharrison:
Google Drive integration is now complete and working as far as I've used it. I've made it match the Dropbox functionality and integration within the app as exactly as possible. I haven't made any tests, but if there is actual interest in getting this merged, please let me know.

@sjhuang26

ngharrison:
Verifying the app through Google and deciding the right permissions would also need to be handled. Since I've been developing it on my own, I have set up an app certificate through my account, have set the permissions to full drive access, and have been accepting them unverified.

ngharrison:
This would close #25.

codemac:
There is also the option to just have a page about how to set up your own app certificate for each user?

This is how many open source apps got around the "security audit" requirements.

marcel-valdez:
Is anyone working on this?

Issue #870, opened 2021-07-15
Customizable timeout when setting Webdav sync


JoeyEremondi:
I have a Nextcloud server running on a 1st generation Raspberry Pi. It is... slow.
When trying to set up WebDAV sync with orgzly, I keep getting a timeout error. I'm fairly certain that the issue is not a problem with the connection, but simply that it is too slow.

Is it possible to configure the timeout, so that I can make it longer to connect to my server? If not, could this be added?
Thanks!

benjaminor:
I have the same issue for my nextcloud server.

Seems like the culprit is OkHttp, which uses default timeouts of 10 seconds (https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html#connectTimeout-long-java.util.concurrent.TimeUnit-).

These timeouts could be changed in https://github.com/orgzly/orgzly-android/blob/ee500503177f7ac598ca872e84f34bad841d111d/app/src/main/java/com/orgzly/android/repos/WebdavRepo.kt by adding the option to set the timeouts of the OkHttpClient.

@nevenz if you want I can create a pull request for this issue?

nevenz:
> if you want I can create a pull request for this issue?

@benjaminor Sure, that would be great, thanks.

nevenz:
I've doubled the values for now. If that still isn't enough, we can add preferences.

yamsu:
Facing the same problem with a nextcloud instance. Perhaps a user defined delay would help.

siriak:
The same problem, had to restart sync like 5 times before it finally managed to sync in time

PR #876, opened 2021-08-14
Introduce custom timeout for server connection (#870)

benjaminor:
Hey, I think this pull request would do the job of enabling a custom timeout for WebDAV repos which
have a problem when syncing because the server is too slow (see #870).

This change would use a user-defined value (in seconds) for all three different timeouts which one can set
with OkHttp (connectTimeout, readTimeout, writeTimeout). I thought this would be better for simplicity.

UI would look like this:
customTimeout

I've tested this with my nextcloud, it seems to be working so far.


nevenz:
I've doubled the values for now, we can see how that goes and add preferences later, thanks.

Issue #790, opened 2020-12-06
Git: During conflicts, each sync creates a new merge branch


amberin:
:heavy_check_mark:  I have searched for existing issues that may be the same as or related to mine.

My intention with #777 was that a merge conflict would lead to Orgzly staying on the first merge branch which could not be merged. Orgzly should then make subsequent commits on this merge branch until it could be merged back into the master/main branch. 

I used to have this working as intended, but apparently I have introduced a bug at some point. Because now, new merge branches are created at each sync occasion, as long as Orgzly is unable to return to main/master.

I expect to be able to fix this soon.

PR #891, opened 2021-09-26
Issue 790, tag solution

amberin:
This solves #790. That issue description does not explain the underlying problem, but I believe the commit messages do.


IvanMalison:
Seems fine, except that I don't think that this is a good use case for a git tag. What a git tag points to should never change over time. Why not just use a git branch?

amberin:
> Seems fine, except that I don't think that this is a good use case for a git tag. What a git tag points to should never change over time. Why not just use a git branch?

The whole point of my last few changes is that we can stay on the "conflict branch" and keep syncing on it, and return to master later.

If we were to use a branch to keep track of the pre-sync commit, while keeping the current (IMO desired) behavior, the work flow would need to be quite cumbersome:
1. Store the current branch name
2. Checkout a new "pre-sync" branch
3. Checkout the previous branch
4. Fetch
5. Attempt merge to master (in case conflicts have been resolved remotely)
6. Start syncing notebooks
7. Delete the pre-sync branch

Step 7 would need to be taken care of by the SyncService class, since only that class knows when syncing is complete. Besides, considering that we create a new merge branch for each notebook with remote changes, I think it would just become confusing with all these different branches.

You are absolutely right that tags should be immutable **if they are shared**, but we never push any tags. This tag is only seen by Orgzly, so I don't really see a problem with moving it around.

IvanMalison:
> If we were to use a branch to keep track of the pre-sync commit

I don't see how using a branch instead of a tag would change anything. They are essentially the same thing (a named ref), except that they are fetched and updated in different ways. If you can use a tag to do this, you can use a branch as a drop in replacement.

Maybe you are worried that a branch somehow needs to be checkout or manipulated with the working tree, while a tag does not. This is absolutely not at all the case. Like I said, branches and tags are essentially the same thing (an alias to a commit), and it should be entirely possible to manipulate a branch in the same way.

IvanMalison:
@amberin See https://en.wikibooks.org/wiki/Git/Advanced

"Both tags and branches point to a commit, they are thus aliases for a specific hash and will save you time by not requiring to type in a hash.

The difference between tags and branches are that a branch always points to the top of a development line and will change when a new commit is pushed whereas a tag will not change. Thus tags are more useful to "tag" a specific version and the tag will then always stay on that version and usually not be changed."

amberin:
You're probably right, I'll look into using a branch instead.

IvanMalison:
> You're probably right, I'll look into using a branch instead.

I think you will probably just want to use the git reset command. IIRC, you can reset a branch to a different commit without even being checked out to that branch.

IvanMalison:
Ah no, I think you just want the equivalent of git branch -f
Soemthing like `git branch -f otherbranch currentbranch`

amberin:
@IvanMalison Thanks for the suggestion, force updating a branch "git branch -f branchname" works great.

@nevenz Feel free to merge at your convenience.

PR #898, opened 2021-10-17
Upgrade to latest JGit

### LINKED ISSUE #892: Git: Upgrade JGit
- I have searched for existing issues that may be the same as or related to mine.

Just creating an issue where we can collect information related to this.

I recently tried switching to the latest JGit, and the first thing I ran into is that the Jsch stuff in GitSSHKeyTransportSetter does not work. But considering our needs, replacing it should not be very hard.

grokII:
Resolves #892

- Upgrade to latest JGIT version
- I've tested it with an emulator and a real device. 
- Executing the `app:lint` gradle task didn't return any errors.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -103,6 +103,7 @@ android {
 
     packagingOptions {
         exclude 'META-INF/DEPENDENCIES'
+        exclude 'plugin.properties'
     }
 }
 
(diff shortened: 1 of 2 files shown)


amberin:
Works on my phone.

Issue #915, opened 2022-01-28
Cannot open file links


michederer:
- I have searched for existing issues that may be the same as or related to mine.

Orgzly 1.8.6 from Play store
Android 11
Redmi Note 10S, M2101K7BNY

Links to files (e.g. [[file:XYZ.txt]]) stopped working for me. 
Dependent on the receiving app they either silently fail or report a "EACCES Permission denied"

A maybe related side note: 
If I try to open some file links (e.g. XYZ.txt) in a browser (e.g. Chrome), an url of the form "content://com.orgzly.fileprovider/external-path-current/XYZ.txt" is shown in the address bar.

michederer:
I changed to fdroid version 1..8.6 => same problem
I changed to fdroid version 1.8.5 => everything works fine

I dont't think that this issue is a duplicate of #913. The problem is related to links to file of the org type "file:" not to notebook files per se. I never had problems with the latter.

crdoconnor:
This is happening to me also. I also dont think it's related to #916

Issue #913, opened 2022-01-26
Local repositories sync is not working(It seems permission denied)


eniehack:
- [x] I have searched for existing issues that may be the same as or related to mine.

Local repositories sync (not Dropbox, enabled auto-sync) is not working. It seems that access right to files are denied.
Also, disabling auto-sync does not change the permission error; checking the permissions(please see pictures below, too) in Orgzly does the same.

Sorry my poor English.
### environment

* Orgzly: latest(1.8.6) from F-Droid
* OS: Android 12
* Phone: Pixel 5a

### pictures

![Screenshot_20220126-022615](https://user-images.githubusercontent.com/32014963/151100184-67333121-1479-43d2-b73a-fe593de3b3d2.png)

![Screenshot_20220126-022619](https://user-images.githubusercontent.com/32014963/151100194-2feff79b-b268-4110-966a-862339a43ac3.png)

### logcat output

```
E MediaProvider: Permission to access file: /storage/emulated/0/orgmodes/notes.org is denied
W System.err: java.io.FileNotFoundException: /storage/emulated/0/orgmodes/notes.org: open failed: EACCES (Permission denied)
W System.err: 	at libcore.io.IoBridge.open(IoBridge.java:575)
W System.err: 	at java.io.FileOutputStream.<init>(FileOutputStream.java:236)
W System.err: 	at java.io.FileOutputStream.<init>(FileOutputStream.java:186)
W System.err: 	at k6.g.b(MiscUtils.java:2)
W System.err: 	at e5.d.c(DirectoryRepo.java:5)
W System.err: 	at q4.y.K1(DataRepository.kt:4)
W System.err: 	at com.orgzly.android.sync.SyncService$a.g(SyncService.kt:17)
W System.err: 	at com.orgzly.android.sync.SyncService$c.a(SyncService.kt:46)
W System.err: 	at com.orgzly.android.sync.SyncService$c.doInBackground(SyncService.kt:1)
W System.err: 	at android.os.AsyncTask$3.call(AsyncTask.java:394)
W System.err: 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
W System.err: 	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:305)
W System.err: 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
[...]

PR #916, opened 2022-01-31
Android R+: request MANAGE_EXTERNAL_STORAGE when creating new git repos

dwoffinden:
I think this also needs requesting on sync or upgrade, in case the app/OS is updated or the permission revoked, but as that's more async I'm not sure where to do it. Just adding the manifest entry makes it possible to manually grant it for now in that case.

An alternative would be to use only app-specific directories permitted under scoped storage, or to perhaps use newer file access APIs to have the user grant fine-grained access? But those'd be larger refactors and this unblocks it for now.

This is a rebase of https://github.com/orgzly/orgzly-android/pull/910, see also https://github.com/orgzly/orgzly-android/issues/24#issuecomment-1025506911

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -16,6 +16,8 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"
         android:maxSdkVersion="28" />
 
+    <uses-permission android:name="android.permission.MANAGE_EXTERNAL_STORAGE" />
+
     <!-- For BroadcastReceiver below -->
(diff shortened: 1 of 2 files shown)


dwoffinden:
I suspect the same permission is also needed for local files: https://github.com/orgzly/orgzly-android/issues/913 & https://github.com/orgzly/orgzly-android/issues/915

mbrgm:
Anything holding this back from being merged?

dwoffinden:
It's a bit of a hack, and only handles new git repos. Ideally we'd do the same thing for local files and on sync/upgrade. But as I said it seems to work for me 🤷

amberin:
@nevenz Please note that Git syncing is completely broken until this has been merged.

Issue #287, opened 2018-01-31
Delete notebook in orgzly when deleted as file (auto-syncing)


iexos:
Every time i delete an org file, it is recreated when opening Orgzly. Instead, I would like to be able to delete a file on another device and have it removed from Orgzly as well.

This is especially important when using Syncthing underneath the file sync, as it can create sync-conflict files which are better resolved (and then deleted) with a proper text editor. Now I have to delete it from Orgzly, and it would make it impossible to use two syncing Orgzly instances.

naught101:
I'm also finding this annoying in Orgzly with syncthing.

It might be good to have an alert when this happens:

"notebook was deleted from repository. Delete from Orgzly too?" Yes/no, and "remember my decision"

Plus something in the settings to reset the "remember my decision" thing. I'm perfectly happy for it to always delete, but it might cause data loss in some cases.

Another option that would help with the syncthing case (and possibly also other cases) would be to have an ignore pattern for org filetypes, e.g. ignore `*.syncthing-conflict-*`.

appetrosyan:
I have an idea for an easy fix. Check if the notebook was ever synced to or from the cloud. If it's present locally, but not on the cloud, ask whether to reupload the local version to the cloud or to remove it.

DrAzraelTod:
Got 4-5 devices that sync files edited by Orgzly (and some text editors)

It's nearly impossible to delete a org-file - like ever

Whenever I'd want to do that, I'll have to delete the file on each device BEFORE one of the other devices synced it's changes back onto my server and re-creates the file.

Alternative Implementation: when a notebook is synced and the file doesn't exist, at least ASK if the file should be re-created.

nevenz:
> Whenever I'd want to do that, I'll have to delete the file on each device BEFORE one of the other devices synced it's changes back onto my server and re-creates the file.

I'm looking into some changes for repos and syncing (for new repo types) and will be having this in mind.

halvorlu:
Any progress on this issue? If not, I might try implementing a solution. My best suggestion is to ask the user whether to delete the file on device, or sync it, as suggested by @DrAzraelTod and @appetrosyan.

nevenz:
> Any progress on this issue?

Not that I know of.

> My best suggestion is to ask the user whether to delete the file on device, or sync it, as suggested by @DrAzraelTod and @appetrosyan.

The problem with asking is that sync is done in the background and can also be started automatically. So user might not be in the app.

[...]

PR #918, opened 2022-02-15
Add option to delete notebook if linked file gets deleted

funsafe-math:
A fix for #287

### DIFF
--- a/app/src/main/res/values-cs-rCZ/strings.xml
+++ b/app/src/main/res/values-cs-rCZ/strings.xml
@@ -590,4 +590,6 @@
     <string name="developer_options">Možnosti pro vývojáře</string>
     <string name="git_repository_type">Typ Git repozitáře</string>
     <string name="in_development">Ve vývoji</string>
+    <string name="pref_title_sync_deletion">Sync deletion</string>
+    <string name="pref_title_summary_sync_deletion">Delete notebook if its deleted from repo</string>
 </resources>
--- a/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
+++ b/app/src/main/java/com/orgzly/android/prefs/AppPreferences.java
@@ -975,6 +975,12 @@ public class AppPreferences {
                 context.getResources().getBoolean(R.bool.pref_default_auto_sync_on_resume));
     }
 
+    public static boolean syncDeletion(Context context) {
+        return getDefaultSharedPreferences(context).getBoolean(
+                context.getResources().getString(R.string.pref_key_sync_deletion),
+                context.getResources().getBoolean(R.bool.pref_default_sync_deletion));
+    }
+
     /*
      * Notes clipboard
      */
--- a/app/src/main/java/com/orgzly/android/sync/SyncService.kt
+++ b/app/src/main/java/com/orgzly/android/sync/SyncService.kt
@@ -307,9 +307,16 @@ class SyncService : Service() {
                             namesake.book.book.id,
                             BookAction.forNow(BookAction.Type.INFO, getString(R.string.canceled)))
 
+                } else if (AppPreferences.syncDeletion(context) &&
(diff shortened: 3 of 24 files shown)


yamanq:
Tested on Android 11, works. Thanks!

brendonscript:
Will this ever be merged @nevenz ? This is a big issue keeping a lot of people from using Orgzly

app/src/main/res/values-cs-rCZ/strings.xml
```diff
@@ -590,4 +590,6 @@
     <string name="developer_options">Možnosti pro vývojáře</string>
     <string name="git_repository_type">Typ Git repozitáře</string>
     <string name="in_development">Ve vývoji</string>
+    <string name="pref_title_sync_deletion">Sync deletion</string>
+    <string name="pref_title_summary_sync_deletion">Delete notebook if its deleted from repo</string>
```
xeruf, line 594:
```suggestion
    <string name="pref_title_summary_sync_deletion">Delete notebook if it is deleted from repo</string>
```
or `it's`, but it is good style to write it out in textual form

same for all the others

xeruf:
Unfortunately some restructuting has happened, raising big conflicts around here :/

Itrekr:
Is there any indication if this is going to get implemented within the foreseeable future? It's really annoying that it's impossible to delete notebooks remotely because orgzly will just re-upload them automatically when it can't find them on the remote server.

mxxivcz:
> it's impossible to delete notebooks remotely

It is now possible to delete a notebook by switching all but one of the devices to plane mode and then deleting the notebook on each of the devices.

Itrekr:
> > it's impossible to delete notebooks remotely
> 
> It is now possible to delete a notebook by switching all but one of the devices to plane mode and then deleting the notebook on each of the devices.

True @ajtakrajta, but that is rather cumbersome if you have to delete a large amount of files. For example, in my workflow I create daily notes which should be disposable. I have a function in Emacs that lets me go through those files and delete them if I no longer need them, and keep them if I do. This can easily be up to 10 files if I do this once every two weeks or so. That means I'll have to carefully select each and every one of those files on my mobile device just so they don't get re-uploaded automatically when I've deleted them on my pc. 

The logical solution for dealing with such conflicts would just be to somehow log the list of note files that orgzly finds in the directory, if, during a sync it can't find one of the notes it previously had, either remove the local note or mark them as a conflict and let the user resolve them.

ugurbolat:
Just a ping.

danielkrajnik:
+1 I hate just repeating what others said, but it's impossible to use Orgzly when it keeps recreating deleted files.

danielkrajnik:
Can someone please merge this? Orgzly is useless without syncing so there is nothing at stake here.

MyFedora:
> Unfortunately some restructuting has happened, raising big conflicts around here :/

No not really.
- Resolve conflicts for translations
- Port changes to `SyncWorker`

I built an apk based on the most recent commit and the changes already, but I'm new to Orgzly, so I didn't verify whether it works yet.

Edit: Yep, works. When I delete org files from a device, Orgzly also deletes the notebook. Here's the patch: [Add-option-to-delete-notebook-if-linked-file-gets-deleted.patch.txt](https://github.com/orgzly/orgzly-android/files/12335646/Add-option-to-delete-notebook-if-linked-file-gets-deleted.patch.txt). Remove `.txt` from the end of the file name.

colonelpanic8:
@tadeuszk733 Would you be interested in trying to merge this into the community fork at 

https://github.com/orgzly-community/orgzly-android

I might consider doing it myself if you think its in a good enough state that it should be merged. WDYT?

colonelpanic8:
@tadeuszk733 as far as I can tell, this doesn't handle the case of two way sync. Is that correct?

PR #931, opened 2022-04-15
Google Drive Integration

sjhuang26:
This PR builds off of #867.

It is currently WIP.


ngharrison:
@sjhuang26 Did you learn anything from working on this? I'm taking a look at it again. Maybe we can finally make it work.

sjhuang26:
@ngharrison I haven't been working on it in a while since I noticed that there's no activity from the maintainer on several other open PRs. If you want to work on the PR, you can rebase my commits (might want to rename them with more descriptive names), and here are some tips:

* Authentication should be working.
* I tried to make the Google Drive and Dropbox implementations as close as possible in the code structure (methods in the same order).
* I'm not sure about this, but I think Dropbox is synchronous while Google Drive _forces_ you to be async (it won't let you run I/O on the main thread). So if the code assumes that I/O is synchronous, this would require refactoring.
* We need to change the path logic. The app only has permission from Google Drive to view/edit directories and files created by the app. (Getting a broader permission requires manual app approval from Google.) Given this:
  * I made a button for the user to create a new directory in the root of My Drive, which the app would own.
  * The user can move this directory, so suppose that the user puts it in `/x/y/z/app_directory/`.
  * Enumerating `/` will yield no results, so the current path logic won't work.
  * I think we should let the user enter directory ID or directory name. Then, we enumerate every directory to find it. Another approach is to have the app enumerate the directories it can see and have the user pick one in a drop-down list.

PR #938, opened 2022-05-20
Apply author/email to the repository, so it also gets used for merges

dwoffinden:
Previously, it was applied only for regular commits, and merges were
attributed to `root`. jgit seems to have no way to specify the comitter,
but it does read from the repo config.


nevenz:
10ec029 has been merged. Let me know if it missed something.

Thanks!

dwoffinden:
Thanks!

Issue #904, opened 2021-11-20
Git: Switch from Jsch to Apache MINA SSHD


amberin:
:heavy_check_mark:  I have searched for existing issues that may be the same as or related to mine.

Jsch is no longer being updated and [the JGit developers are telling users to switch to the Apache MINA SSHD library](https://www.eclipse.org/forums/index.php?t=msg&th=1106995&goto=1838158&#msg_1838158).

I recently upgraded to OpenSSH 8.6 on my workstation, and I can no longer connect to it from Orgzly's current Jsch SSH client, because the latter does not support 256-bit signing algorithms for RSA keys (at least not from what I can tell). (I created a new RSA key which works fine from other SSH clients, but not from Orgzly/Jsch.)

Since Jsch also only supports RSA keys, this will soon cause problems for all Git+SSH users. (The Apache library supports many modern key types.)

colonelpanic8:
@amberin did you ever take a swing at this? do you have a sense of how hard its going to be?

PR #948, opened 2022-07-19
Git via SSH: Replace Jsch transport with Apache

amberin:
Apache MINA SSHD is now the standard SSH transport library for JGit. It supports more modern key algorithms than Jsch. This is my implementation attempt. It uses a .ssh directory in the app's private storage for storing "known hosts" information.

I hope to look into generating the SSH private key on the device shortly.

I have added a notification prompt to the user upon new or changed SSH server host key. I am probably doing this completely wrong, but it seems to work.

N.B. Currently requires API >= 26, but I suspect that would be easy to fix for someone who actually knows Java and Android.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -192,7 +192,11 @@ dependencies {
     implementation "io.github.rburgst:okhttp-digest:$versions.okhttp_digest"
 
     implementation "org.eclipse.jgit:org.eclipse.jgit:$versions.jgit"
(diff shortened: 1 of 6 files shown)


amberin:
Intended to solve #904

amberin:
Rebased on current master.

amberin:
Closing this in favor of #1037.

PR #1021, opened 2022-12-30
Git sync: Push current branch to remote if it is empty

amberin:
This allows the user to easily set up a new "origin" remote (by initalizing an empty repository) if the current one is somehow lost.

A Git repo which is synced between two locations should have a degree of redundancy. Without this code, there is no convenient way to sync the repo contents using Orgzly as the source.


amberin:
Closing this in favor of #1037.

PR #1022, opened 2023-01-02
Monochrome vector launcher icon for adaptive-icons support

markhepburn:
This is primarily an Android 13 / "Material You" detail ([reference](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive)).  The monochrome option is just the existing foreground icon with extraneous detail tidied up.

Before:

After:


cbkerr:
Fixed in https://github.com/orgzly-revived/orgzly-android-revived/pull/18

markhepburn:
Thanks!

PR #1040, opened 2023-05-15
Need apk artifact with time clocking enabled 

alokehpdev:
Time clocking is implemented with 1.8.11 but apk is nowhere to be found. Every other artifact has expired. Just wanted an apk actually with latest time clocking merged.


cbkerr:
@alokehpdev, these are available in Orgzly Revived, a community fork.

https://github.com/orgzly-revived/orgzly-android-revived/releases

PR #1054, opened 2024-03-02
Add support for TLSv1.2 for WebDAV in Android 4.x

deuill:
**NOTE:** I am aware that this project is currently unmaintained, and even so this change would not make it in as support for pre-5.0 Android versions was dropped in version 1.8.6.

This PR is intended with pushing a patch and APK I had to create to get WebDAV working with modern TLS on an old Android device (or, more specifically, a Blackberry Q10 running its Android emulator). Other people might find it useful. The base of this PR is the `v1.8.5` tag.

Android 4.1 to 5.0 have support for TLSv1.2 in creating encrypted sockets, but this is (for some reason) disabled by default. This commit adds conditional support for using a TLS socket override with added support for TLSv1.2 in these SDK build versions, in making WebDAV connections.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ buildscript {
 
     versions.glide = '4.10.0'
 
-    versions.sardine = '0.5'
+    versions.sardine = '0.4'
 
     versions.jgit = '4.4.1.201607150455-r'
 
--- a/app/src/main/java/com/orgzly/android/repos/WebdavRepo.kt
+++ b/app/src/main/java/com/orgzly/android/repos/WebdavRepo.kt
@@ -1,18 +1,23 @@
 package com.orgzly.android.repos
 
 import android.net.Uri
+import android.os.Build
 import com.orgzly.android.BookName
+import com.orgzly.android.util.TLSSocketFactory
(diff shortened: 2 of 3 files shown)


deuill:
Attaching a self-signed APK, built against SDK version 18 (Android 4.3 Jelly Bean). Hope it's useful to others!

[app-fdroid-release.apk.zip](https://github.com/orgzly/orgzly-android/files/14467140/app-fdroid-release.apk.zip)

build.gradle
```diff
@@ -59,7 +59,7 @@ buildscript {
 
     versions.glide = '4.10.0'
 
-    versions.sardine = '0.5'
```
deuill, line 62: Version 0.5 brings an upgrade to OKHTTP 4.0, which breaks connections on older Android versions.

amberin:
There is an active fork of this project: https://github.com/orgzly-revived/orgzly-android-revived

deuill:
> There is an active fork of this project: https://github.com/orgzly-revived/orgzly-android-revived

The revived version doesn't have support for SDK targets required here (essentially Android 4.3 and below).

Issue #1059, opened 2024-09-22
Search expression to match only "leaf" notes


bobrippling:
- [X] I have searched for existing issues that may be the same as or related to mine.

If I have notes such as:
```org
* Shopping :shopping:
** Bread
** Veg
*** Carrots
*** Cabbage
```

Where I've grouped some items, I'd like to find all "leaf" items, and ignore the groups. In other words, items without children, so in this case, `Bread`, `Carrots` and `Cabbage`.

Would a change to add a `leaf` search expression be accepted? If so I can create a PR. It would be something like:

`t.shopping leaf`
or
`t.shopping hierarchy.leaf` / `t.shopping h.leaf` if we wanted to avoid the "leaf" bareword being by itself

TheAttentionSeeker5050:
I'm making progress on implementing the leaf note search feature. A summary of my work up to date.

- I've introduced a recursive CTE to fetch child elements of live data notes after applying other filters.
- Added a new boolean property `searchLeafNotes` to the Options data class. This will act as a trigger for the new search option in DataRepository.

Here's how it works (using "h.leaf" in the search bar):
1. For tags: Get all "t.shopping" IDs, then fetch all leaf notes for "t.shopping h.leaf"
2. For title/content: Get list item IDs with "shopping" in title or content, then retrieve all leaf children of those IDs

Summary of fix list:
- Tag leaf note searches are working as expected, which is good
- Title/content searches aren't quite there yet (still returning non-leaf results)
- During debugging, I confirmed the first query (getting filtered note IDs) works in both cases

The raw SQL runs fine in SQLite Workbench, but I need to refine the Android methods to ensure it works for most, if not all, search filters.

I'll submit a PR once all my tests are passing.

bobrippling:
Oh great stuff, thank you! That was a quick turn around to a PR too, thanks for summarising here. I've a few questions but I'll pop them on the PR

amberin:
@bobrippling @TheAttentionSeeker5050 This project was abandoned in 2022. There is a fork here: https://github.com/orgzly-revived/orgzly-android-revived

bobrippling:
Ah great - thanks @amberin!

PR #1067, opened 2025-03-18
Feature/search leaf nodes

TheAttentionSeeker5050:
Implementation for feature: Search expression to match only "leaf" notes #1059

Using two queries to get only the leaf notes when the search input includes h.leaf (leaf hierarchy).

The first SQL query works as usual. If it's a leaf search, it's saved in the Options property called searchLeafNotes.

Inside selectNotesFromQueryLiveData, once this property is checked and we get the NoteView LiveData array based on other criteria, the process maps them into a list of Long note IDs. Then, another SQL query runs to recursively find child notes from the original search. Finally, it filters out any notes that are parents of another note in the results, leaving only the leaf notes.

Tested with the following inputs: 
- t.tag2 h.leaf
- fold h.leaf 
- i.todo h.leaf


app/src/main/java/com/orgzly/android/query/user/DottedQueryParser.kt
```diff
@@ -113,6 +113,10 @@ open class DottedQueryParser : QueryParser() {
             OptionMatch("""^ad\.(\d+)$""") { match, options ->
                 val days = match.groupValues[1].toInt()
                 if (days > 0) options.copy(agendaDays = days) else null
+            },
+            OptionMatch("""^(\.)?h\.(.+)""") { match, options ->
+                if (match.groupValues[2].lowercase() == HyerarchyType.LEAF.toString().lowercase())
```
bobrippling: Perhaps the file that defines `HyerarchyType` is missing?

bobrippling (commented):
Nice work! I wonder - what will happen when the leaf query isn't at the top level, for example: `t.work or (b.gifts and h.leaf)` ?

app/src/main/java/com/orgzly/android/query/user/DottedQueryParser.kt
TheAttentionSeeker5050: I will check this out Bob.

TheAttentionSeeker5050: Also, the HierarchyType enum class was not added into the Git commit. As it was part of a previous iteration to solve this problem. 

I am working on it

TheAttentionSeeker5050: Added HierarchyType and fixed its name typo (before, it was spelled incorrectly as HyerarchyType)

TheAttentionSeeker5050:
> t.work

Can you please explain this more in depth? I only set up the leaf option as of this date on hierarchy type. So the query with this option on will make all the query then perform the search leaf notes and filter out the parents. The order of the input doesn't affect the result as far as I am aware. Example: `fold h.leaf`, `fold and h.leaf`, `h.leaf fold`, `fold (h.leaf)`, `(fold) h.leaf `, `(fold h.leaf)` all return the same result.

Edit: The leaf notes means that it will only return the very end children of the search query, all the parents and intermediate children do not make part of the end result. They are filtered out.

app/src/main/java/com/orgzly/android/query/user/DottedQueryParser.kt
bobrippling: Nice, thanks! I guess the filename just needs the typo change too?

TheAttentionSeeker5050: I fixed the typo

TheAttentionSeeker5050: Youre right the file name sorry I will fix it ASAP

bobrippling:
> > t.work
> 
> Can you please explain this more in depth? I only set up the leaf option as of this date on hierarchy type. So the query with this option on will make all the query then perform the search leaf notes and filter out the parents. The order of the input doesn't affect the result as far as I am aware. Example: `fold h.leaf`, `fold and h.leaf`, `h.leaf fold`, `fold (h.leaf)`, `(fold) h.leaf `, `(fold h.leaf)` all return the same result.

Sure - what I mean is that the leaf part is more a condition than an option, so we want it to behave like [tag](https://github.com/orgzly/orgzly-android/blob/271b6fc8048c13b343b80dbcd8d64759cf6e2ca6/app/src/main/java/com/orgzly/android/query/user/BasicQueryParser.kt#L34-L36=), or [state](https://github.com/orgzly/orgzly-android/blob/271b6fc8048c13b343b80dbcd8d64759cf6e2ca6/app/src/main/java/com/orgzly/android/query/user/BasicQueryParser.kt#L17-L19=) for example.

This would mean we can then look for notes which are either:
- tagged with "work" (while not necessarily a leaf note)
- leaf-notes which are tagged with "gift"

(in the above example, `t.work or (b.gifts and h.leaf)`), rather than it being an all-or-nothing option, if you see what I mean?

app/src/main/java/com/orgzly/android/query/user/DottedQueryParser.kt
bobrippling: No worries, thanks!

TheAttentionSeeker5050:
> > > t.work
> > 
> > 
> > Can you please explain this more in depth? I only set up the leaf option as of this date on hierarchy type. So the query with this option on will make all the query then perform the search leaf notes and filter out the parents. The order of the input doesn't affect the result as far as I am aware. Example: `fold h.leaf`, `fold and h.leaf`, `h.leaf fold`, `fold (h.leaf)`, `(fold) h.leaf `, `(fold h.leaf)` all return the same result.
> 
> Sure - what I mean is that the leaf part is more a condition than an option, so we want it to behave like [tag](https://github.com/orgzly/orgzly-android/blob/271b6fc8048c13b343b80dbcd8d64759cf6e2ca6/app/src/main/java/com/orgzly/android/query/user/BasicQueryParser.kt#L34-L36=), or [state](https://github.com/orgzly/orgzly-android/blob/271b6fc8048c13b343b80dbcd8d64759cf6e2ca6/app/src/main/java/com/orgzly/android/query/user/BasicQueryParser.kt#L17-L19=) for example.
> 
> This would mean we can then look for notes which are either:
> 
> * tagged with "work" (while not necessarily a leaf note)
> * leaf-notes which are tagged with "gift"
> 
> (in the above example, `t.work or (b.gifts and h.leaf)`), rather than it being an all-or-nothing option, if you see what I mean?

Now I understand. I will have a look at this case.

app/src/main/java/com/orgzly/android/query/user/DottedQueryParser.kt
TheAttentionSeeker5050: File renamed

amberin:
@TheAttentionSeeker5050  @bobrippling Hi guys, this project has been abandoned for three years now. There is an active fork here: https://github.com/orgzly-revived/orgzly-android-revived

Feel free to open a PR!

bobrippling:
Thanks @amberin! For reference / anyone following, the corresponding orgzly-revived issue is here: orgzly-revived/orgzly-android-revived#653

